PR #2, opened 2023-06-20
Permission Take 1

willronchetti:
- Implements a permissions scheme centered around two item types: `consortium` and `submission_center` associations on types
- `submission_centers` is intended to allow access only to the submitting center when types are in review (ie: ongoing submission)
- `consortiums` is intended to allow access to all consortium level users for data that is "released to consortium" (status values still in flux) 
- Some significant tests for the above with default statuses


src/encoded/tests/test_permissions.py
```diff
@@ -8,59 +8,142 @@ def submission_center_user_app(testapp, test_submission_center, smaht_gcc_user):
...
+            item_uuid=uuid
+        )
 
-        # patch the file status so it has no submission_center, user should now no longer see
-        testapp.patch_json(f'/{file["uuid"]}?delete_fields=submission_centers', {}, status=200)
-        submission_center_user_app.get(f'/{file["uuid"]}', status=403)
-        anontestapp.get(f'/{file["uuid"]}', status=403)
-        testapp.get(f'/{file["uuid"]}', status=200)
+        # patch the file status, so it has no submission_center therefore user should no longer see
+        testapp.patch_json(f'/{submission_center_file["uuid"]}?delete_fields=submission_centers', {}, status=200)
```
netsettler, line 86: why patch and not just use a different submission center? won't it be faster and possibly more robust? what mechansim puts it back?

src/encoded/types/document.py
```diff
@@ -16,4 +16,5 @@
         'description': 'Listing of Documents',
     })
 class Document(SMAHTItem, CoreDocument):
```
netsettler, line 18: I wonder if we should just use a class like SmahtItem, though I guess that's not a whole lot more readable. Maybe ignore me...

src/encoded/types/access_key.py
```diff
@@ -18,4 +27,39 @@
...
 class AccessKey(SMAHTItem, SnovaultAccessKey):
     """AccessKey class."""
+    ACCESS_KEY_EXPIRATION_TIME = 90  # days
+    item_type = 'access_key'
+    name_key = 'access_key_id'
     schema = mixin_smaht_permission_types(SNOVAULT_ACCESS_KEY_SCHEMA)
+    embedded_list = []
+
+    STATUS_ACL = {
+        'current': [(Allow, 'role.owner', ['view', 'edit'])] + ONLY_ADMIN_VIEW_ACL,
```
netsettler, line 37: Did we talk about naming this last item something like `ELSE_ONLY_ADMIN_VIEW_ACL` or `OTHERWISE_ONLY_ADMIN_VIEW_ACL` to encourage it being last. It's in the right position here but it might be easier to spot it in the wrong position if it had a better name.

src/encoded/types/access_key.py
```diff
@@ -18,4 +27,39 @@
...
+    class Collection(SMAHTItem.Collection):
+        pass
+
+
+@view_config(context=AccessKey.Collection, request_method='POST',
+             permission='add',
+             validators=[validate_item_content_post])
+@debug_log
+def access_key_add(context, request):
+    return sno_access_key_add(context, request)
```
netsettler, line 50: Guessing this has to be done this way rather than by just using include from snovault so it goes through the right permissioning?

src/encoded/tests/test_types_submission_center.py
```diff
@@ -1,3 +1,3 @@
 def test_types_submission_center(testapp, test_submission_center):
     """ Tests we can post/get a consortium item """
-    testapp.get(f'/submission-center/{test_submission_center["name"]}')
```
netsettler, line 3: A random place to mention this, but ... do we want to have version numbers on this new protocol in case we later have to change it?

src/encoded/types/base.py
```diff
@@ -32,47 +32,56 @@ def mixin_smaht_permission_types(schema: dict) -> dict:
...
+            PRINT(f'DEBUG_PERMISSIONS: returning {ALLOW_SUBMISSION_CENTER_CREATE_ACL} for {self.type_info.name}')
+            self.__acl__ = ALLOW_SUBMISSION_CENTER_CREATE_ACL
+        else:
+            self.__acl__ = ONLY_ADMIN_VIEW_ACL
+            PRINT(f'DEBUG_PERMISSIONS: using admin acl for {self.type_info.name}')
 
 
 @abstract_collection(
-    name='smaht-items',
+    name='items',
```
netsettler, line 66: I don't know how I feel about renaming this but not renaming SMaHTItem. Can you elaborate on how I'm supposed to think about why that is? Maybe even document in a comment someplace, though this might not be that place.

src/encoded/types/user.py
```diff
@@ -17,7 +46,7 @@
 )
 class User(SMAHTItem, SnovaultUser):
     """ Overridden user class, adding the Submission Center and Consortium attribution """
-    item_type = 'smaht-user'
+    item_type = 'user'
```
netsettler, line 49: Here you renamed both the class and the item type, unlike elsewhere that you did not.

src/encoded/types/user.py
```diff
@@ -7,6 +7,35 @@
 SNOVAULT_USER_SCHEMA = deepcopy(SnovaultUser.schema)
 
 
+# def mixin_smaht_permission_types(schema: dict) -> dict:
```
netsettler, line 10: What's the status of this code?

netsettler (approved):
A lot of questions. Nothing that is an obvious blocker, but please do check out the marked issues.

drio18 (commented):
Left a few smaller comments, but, overall, I'm a little confused about the necessity of the Consortium item and its associated permissions. Will there only be 1 consortium (SMaHT)? If not, what others are we expecting? Is the plan to associate a consortium with all users or at least all those not associated with a submission center, and thus the consortium essentially is serving as authenticated user permissions? It would be good to have a quick reference doc for all of this in the portal somewhere to understand the logic at a high-level.

src/encoded/schemas/mixins.json
```diff
@@ -471,5 +471,25 @@
...
+        "submission_centers": {
+            "title": "Submission Centers",
+            "description": "Submission Centers associated with this item.",
+            "type": "array",
+            "items": {
+                "linkTo": "SubmissionCenter"
+            },
+            "serverDefault": "user_submission_centers"
+        },
+        "consortiums": {
```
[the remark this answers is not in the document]
willronchetti: I think consortia sounds better personally so probably will just change it to that.

src/encoded/tests/test_permissions.py
```diff
@@ -8,59 +8,142 @@ def submission_center_user_app(testapp, test_submission_center, smaht_gcc_user):
...
     }, status=201).json['@graph'][0]
+
+
+@pytest.fixture
+def submission_center_file(testapp, fastq_format, test_submission_center):
     item = {
-        'file_format': res['uuid'],
+        'file_format': fastq_format['uuid'],
         'md5sum': '00000000000000000000000000000000',
         'filename': 'my.fastq.gz',
```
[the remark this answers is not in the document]
willronchetti: The intent here is to test data model driven permissions, not anything specific about the file itself. This could actually be tested with other types but I happened to use a file type for a more complex case.

src/encoded/tests/test_permissions.py
willronchetti: All datafixtures are function scoped so the changes in the test are wiped before the next test. Even with a small # of tests here they seem to go very fast per test even when building multiple data model elements so have preferred this over the workbook in the meantime.

src/encoded/types/user.py
willronchetti: Removed

src/encoded/types/base.py
```diff
@@ -32,47 +32,56 @@ def mixin_smaht_permission_types(schema: dict) -> dict:
...
     # Ie: if an item status = public, then the ACL ALLOW_EVERYONE_VIEW applies to its permissions,
     # so anyone (even unauthenticated users) can view it
-    STATUS_ACL = {
+    SUBMISSION_CENTER_STATUS_ACL = {
         'shared': ALLOW_CONSORTIUM_MEMBER_VIEW_ACL,
         'obsolete': ALLOW_CONSORTIUM_MEMBER_VIEW_ACL,
         'current': ALLOW_CONSORTIUM_MEMBER_VIEW_ACL,
         'inactive': ALLOW_SUBMISSION_CENTER_MEMBER_VIEW_ACL,
-        'in review': ALLOW_SUBMISSION_CENTER_MEMBER_EDIT_ACL,
+        'in review': ALLOW_SUBMISSION_CENTER_MEMBER_VIEW_ACL,
```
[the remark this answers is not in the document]
willronchetti: Yes this is a mistake, will revert back to edit

src/encoded/tests/data/master-inserts/submission_center.json
```diff
@@ -0,0 +1,8 @@
+[
+  {
+    "name": "Park Lab Submission Center",
+    "title": "Park Lab Submission Center",
```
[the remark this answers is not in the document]
netsettler: I marked thumbs up, but I wanted to clarify that's me agreeing with doug, not me saying I'll fix it.

src/encoded/types/acl.py
```diff
@@ -12,7 +12,7 @@
 #     2. __acl__ method within Collection class
 #           This controls who can read/edit metadata items part of the collection class
 #     3. __ac_local_roles__ method within Collection class
-#           This
+#           This controls
```
netsettler, line 15: Some reason this sentence is still not complete?

src/encoded/types/base.py
```diff
@@ -31,13 +31,15 @@ def mixin_smaht_permission_types(schema: dict) -> dict:
...
         },
         'serverDefault': 'user_submission_centers'
     }
     schema['properties']['consortiums'] = {
         'type': 'array',
         'items': {
+            'type': 'string',
             'linkTo': 'Consortium'
         },
         'serverDefault': 'user_consortiums'
```
netsettler, line 45: Should this be called `user_consortiums`? Does it make one or several?

dmichaels-harvard (approved):

PR #3, opened 2023-06-26
Updated dcicsnovault to 8.2.0.1b40.

dmichaels-harvard:


willronchetti (approved):
I would bump version then merge so this can get in and things can be more stable

PR #4, opened 2023-06-28
Deployment

willronchetti:
- Updates various deployment components to ensure a baseline level of functionality so the portal deploys


dmichaels-harvard (approved):

src/encoded/__init__.py
```diff
@@ -22,7 +22,7 @@
 from snovault.elasticsearch.interfaces import INVALIDATION_SCOPE_ENABLED
 from dcicutils.misc_utils import VirtualApp
 from .appdefs import APP_VERSION_REGISTRY_KEY
-from snovault.loadxl import load_all
+# from snovault.loadxl import load_all
```
netsettler, line 25: So this was imported but not used? Interesting. I guess that was so it could be exported from here?

I mostly am not a fan of there ever being code in `__init__.py` at all, and maybe this is partly the reason. People use these as meaning "import all the useful stuff" and yet if you're going to do implementation work as well, then it has the problem of potentially creating new stuff that the included stuff comes and that's a huge circularity risk. I'd really rather we adopt a style of literally only doing imports from a lower package in one of these files and never in those lower packages importing from here, and moving all the implementation stuff to a, I dunno, let's call it a `module_setup.py` file, if the `util.py`† doesn't work, and then importing relevant bits from there.

†or `utils.py` (we should really re-standardize the spelling of that across portal repos, after I started to change it and never completed, so left it in an incomplete state)

src/encoded/__init__.py
```diff
@@ -141,13 +141,13 @@ def robots_txt(request):
...
 
 
-def load_workbook(app, workbook_filename, docsdir):
-    environ = {
-        'HTTP_ACCEPT': 'application/json',
-        'REMOTE_USER': 'IMPORT',
-    }
-    testapp = VirtualApp(app, environ)
-    load_all(testapp, workbook_filename, docsdir)
+# def load_workbook(app, workbook_filename, docsdir):
```
netsettler, line 144: In case this function isn't just totally on it's way out...

If this is useful, maybe it should move to either util.py or to loadxl.py itself.

Also, speaking of utility functions, there is a function that makes a testapp of the appropriate kind, and people shouldn't be composing them raw like this in my view.

PR #6, opened 2023-07-09
Data model additions

drio18:
Here, we begin to add some core item types for SMaHT not present in encoded-core, particularly those that will be submitted to us.

The overall philosophy for the current design centered around the submission process (flat items, overall top down with many rev-links for collecting items), limiting separate types (via implementing "categories" and making greater use of schema validation as needed), and consolidation of property names for consistency across item types.

A poorly drawn version of the current idea can be seen [here](https://docs.google.com/drawings/d/10-hxp8h4Rz3lQ41yT9gAICczIT0lcqLg-XPpkDXkFdY/edit?usp=sharing). Note that the item types are not 1-to-1 with the drawing yet.

A few specific comments to head off potential questions:

- Majority of expected properties not included here. Looking to put in the skeleton with this PR and to finalize the design, then properties can quickly follow once confirmed.
- `submitter_id` is intended for use with submitted items for users to refer to them uniquely per submission center, an overlapping idea with (and intended replacement for) aliases, though the latter can continue to be used internally if helpful.
- `identifier` is intended to be the definitive item "barcode", akin to [the idea for TCGA](https://docs.gdc.cancer.gov/Encyclopedia/pages/TCGA_Barcode/), if we're heading down that path for certain items.

Potential sources of feedback:

- How does everyone feel about maintaining minimal types and going the category route as done here? It will result in properties that are not applicable to all items compared to the distinct types approach, but it simplifies the data model in some sense.
- Should we reduce file types to 1 generic type as well with categories as above or stick with the traditional troika (submitted, processed, reference)?

Questions that remain to be resolved:

- Utility of an analysis item? May be useful to separate out pipelines and output files more cleanly, particularly if there are multiple pipeline sets run in time, at the cost of an additional item type.
- Creating schemas for items from encoded-core once ability to reference them incorporated.


src/encoded/schemas/aliquot.json
```diff
@@ -0,0 +1,86 @@
+{
+    "title": "Analyte",
+    "id": "/profiles/analyte.json",
```
willronchetti, line 3: Names here don't match?

src/encoded/schemas/biosample.json
```diff
@@ -0,0 +1,114 @@
...
+            "title": "Biosources",
+            "type": "array",
+            "uniqueItems": true,
+            "minItems": 1,
+            "items": {
+                "title": "Biosource",
+                "type": "string",
+                "linkTo": "Biosource"
+            }
+        },
```
willronchetti, line 94: Could a biosample come from multiple biosources? How would that work? Mixed sample?

src/encoded/types/acl.py
```diff
@@ -15,10 +13,10 @@
...
 #           ie: consortia, submission center presence indicates permissions (roles) should be given
 #           to a user
-CONSORTIUM_MEMBER_CREATE = 'role.consortium_member_create'
-SUBMISSION_CENTER_MEMBER_CREATE = 'role.submission_center_member_create'
-CONSORTIUM_MEMBER_RW = 'role.consortium_member_rw'
-SUBMISSION_CENTER_RW = 'role.submission_center_member_rw'
+CONSORTIUM_MEMBER_CREATE = "role.consortium_member_create"
+SUBMISSION_CENTER_MEMBER_CREATE = "role.submission_center_member_create"
+CONSORTIUM_MEMBER_RW = "role.consortium_member_rw"
+SUBMISSION_CENTER_RW = "role.submission_center_member_rw"
```
willronchetti, line 20: What linter are you using? We should standardize this as a make target so formatting can be done automatically so we don't get large diffs across the repo like this that are pointless.

src/encoded/types/aliquot.py
```diff
@@ -0,0 +1,46 @@
+from typing import Iterable, List, Optional, Union
+
+from pyramid.request import Request
+from snovault import calculated_property, collection
+
+from .base import SMAHTItem, load_smaht_schema
```
willronchetti, line 6: Will break upon merging master in, see other types files to correct this import

src/encoded/types/analyte.py
```diff
@@ -0,0 +1,46 @@
...
+from .utils import CENTER_SUBMITTER_IDS_SCHEMA, get_unique_center_submitter_ids
+
+
+@collection(
+    name="analytes",
+    unique_key="accession",
+    properties={"title": "Analytes", "description": "Listing of analytes"},
+)
+class Analyte(SMAHTItem):
+    item_type = "analyte"
```
willronchetti, line 16: Please docstring all new types so we have some basic info in code on what they are/what important relationships exist ie revlinks especially.

src/encoded/types/experiment.py
```diff
@@ -0,0 +1,28 @@
...
+    @calculated_property(schema=CENTER_SUBMITTER_IDS_SCHEMA)
+    def center_submitter_ids(
+        self,
+        request: Request,
+        submitter_id: Optional[str] = None,
+        submission_centers: Optional[Iterable[str]] = None,
+    ) -> Union[List[str], None]:
+        if submitter_id and submission_centers:
+            return get_unique_center_submitter_ids(submitter_id, submission_centers, request)
+        return
```
willronchetti, line 28: This is repeated throughout, maybe should be a calc prop on the base item?

willronchetti (approved):
These changes look fine but if you can add install/make target instructions for your linter it will save us some significant diffs in the future. You may encounter significant merge conflicts when merging master in though due to changes in base.py (and the linter).

aschroed (commented):
I'm still going through this so not ready to comment yet but one question I have is the concept of 'Consortium' - this may be more a question for @willronchetti but I see that a schema has been added for this as an item type and it looks to be heavily integrated into permissions and upcoming changes to submitCGAP so I wanted to understand the thoughts regarding multiple 'Consortia' (if anticipated) and why the Item?  Or does Consortium mean something else in this context.

src/encoded/schemas/ontology_term.json
```diff
@@ -0,0 +1,70 @@
...
+            "$ref": "mixins.json#/uuid"
+        }
+    ],
+    "type": "object",
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "accession": {
+            "accessionType": "OT"
```
aschroed, line 65: Do you really want to accession Ontology Terms or Ontologies?  I see you are accessioning many of the items in your schemas.  Ontology maintainers are generally fairly good at having a stable ID and uri for ontology terms and the ontology itself.  Will you reaccession with every term update?  Well behaved ontologies have mechanisms for obsoletion as opposed to updates and managing this within the portal via accessioning seems potential overkill and potential additional work to maintain.

At least in 4DN accessions are meant to be permanent identifiers that will always point to the same item - even with the replacement mechanism that we use if a file needs to be retired you that provide a redirect if the old accession is used the old item is still accessible.

We do not have versioning though we have definitely considered it as a way to allow updates without needing to mint new accessions - something to consider especially for protocol and quality metric items?

PR #7, opened 2023-07-19
Add More Tests

willronchetti:
- Improve testing by porting relevant tests as needed
- Changes to accommodate working search/other tests
- Allow testing with ES in GA with smaht-development credentials
- Build Docker as part of GA

### DIFF
--- a/src/encoded/tests/test_schemas.py
+++ b/src/encoded/tests/test_schemas.py
@@ -1,0 +1,196 @@
+import pytest
+import re
+
+from pkg_resources import resource_listdir
+from snovault import COLLECTIONS, TYPES
+from snovault.schema_utils import load_schema
+
+
+pytestmark = [pytest.mark.setone, pytest.mark.working, pytest.mark.schema, pytest.mark.indexing]
+
+
+SCHEMA_FILES = [
+    f for f in resource_listdir('encoded', 'schemas')
+    if f.endswith('.json') and 'embeds' not in f
+]
+
+
+def pluralize(name):
+    """ This is a special function used to pluralize in a somewhat random way, but kept for legacy reasons... """
+    name = name.replace("_", "-")
+    # deal with a few special cases explicitly
+    specials = ["file", "quality-metric", "summary-statistic", "workflow-run", "consortium"]
+    for sp in specials:
+        if name.startswith(sp) and re.search("-(set|flag|format|type)", name) is None:
+            return name.replace(sp, sp + "s")
+        elif name.startswith(sp) and re.search("setting", name):
+            return name.replace(sp, sp + "s")
+        elif name == "consortium":  # special case for consortium
+            return name
+    # otherwise just add 's/es/ies'
+    if name.endswith("ly"):
+        return name[:-1] + "ies"
+    if name.endswith("sis"):
+        return name[:-2] + "es"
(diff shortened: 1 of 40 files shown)


src/encoded/tests/verifier.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def verifier(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        print("running tests " + func.__name__)
+        try:
+            res = func(*args, **kwargs)
+        except Exception as e:
+            print("test failed with exception " + str(e.args))
```
netsettler, line 16: Most people just use str(e). Is there some reason this is better?

src/encoded/tests/verifier.py
```diff
@@ -0,0 +1,113 @@
...
+    db_item = indexer_testapp.get("/" + item_uuid + "/?datastore=database").follow(status=200).json
+    ensure_basic_data(db_item, item_type)
+
+
+@verifier
+def verify_profile(item_type, indexer_testapp):
+    # is this something we actually know about?
+    profile = indexer_testapp.get("/profiles/" + item_type + ".json").json
+    assert profile
+    item_type_camel = profile['id'].strip('.json').split('/')[-1]
```
netsettler, line 54: could we have a separate (and tested) function that does this complicated dance so i know what it's achieving?

src/encoded/tests/verifier.py
```diff
@@ -0,0 +1,113 @@
...
+    ensure_basic_data(es_item, item_type)
+    return es_item, item_type
+
+
+@verifier
+def verify_get_by_accession(es_item, item_type, indexer_testapp):
+    # get by accession
+    accession = es_item.get('accession')
+    if accession:  # some items don't have acessions
+        item_by_accession = indexer_testapp.get("/" + accession).follow(status=200).json
```
netsettler, line 38: seems like this idiomatic use should have a function name, too. we sure do it a lot.

src/encoded/tests/verifier.py
```diff
@@ -0,0 +1,113 @@
+from dcicutils.misc_utils import ignored
+from functools import wraps
+from snovault import TYPES
+from .test_create_mapping import test_create_mapping_correctly_maps_embeds
+#from .test_embedding import test_add_default_embeds, test_manual_embeds
```
netsettler, line 5: pep8 is going to beat you up about lack of space character after '#'

src/encoded/tests/test_schemas.py
```diff
@@ -0,0 +1,196 @@
...
+pytestmark = [pytest.mark.setone, pytest.mark.working, pytest.mark.schema, pytest.mark.indexing]
+
+
+SCHEMA_FILES = [
+    f for f in resource_listdir('encoded', 'schemas')
+    if f.endswith('.json') and 'embeds' not in f
+]
+
+
+def pluralize(name):
```
netsettler, line 18: I would like to think this does the same kinds of things as dcicutils.lang_utils.pluralize, or that if it does something that function doesn't that we fix that.

netsettler (commented):
This looks OK to me.

netsettler (approved):
This looks OK to me.

PR #13, opened 2023-08-22
Submitr/ingestion related work

dmichaels-harvard:
* Adding ingestion processor.
  * Added ingestion_processor.py; basically a stub for now, using Kent's sheet_utils.
  * Added generate-local-access-key script (from snovault) to pyproject.toml.
  * Changed metadata_bundles_bucket to smaht-production-application-metadata-bundles in development.ini.template


willronchetti (approved):
All looks reasonable

src/encoded/ingestion/ingestion_processors.py
```diff
@@ -0,0 +1,42 @@
...
+        validate_data_problems = validate_data_against_schemas(data, portal_vapp=submission.portal_vapp)
+        if validate_data_problems:
+            validate_data_summary = summarize_validate_data_problems(validate_data_problems, submission)
+            submission.record_results(validate_data_problems, validate_data_summary)
+            # If there are data validation problems then trigger an exception so that a traceback.txt
+            # file gets written to the S3 ingestion submission bucket to indicate that there is an
+            # error; this is done in snovault.types.ingestion.SubmissionFolio.processing_context.
+            # TODO: I *think* this is the right thing to do; it is an exceptional
+            # situation but we just happened to have caught it programmatically.
+            raise Exception(validate_data_problems)
```
willronchetti, line 33: So long as this results in `errors` being appropriately populated on the ingestion submission, sounds good

src/encoded/ingestion/loadxl_extensions.py
```diff
@@ -0,0 +1,79 @@
...
+    object in the Portal database (see SubmissionFolio.record_results); this is what will
+    get displayed, by default, by the submitr tool when it detects processing has completed.
+    """
+    return [
+        f"Successful ingestion summary:",
+        f"In File: {submission.data_file_name}",
+        f"S3 File: {submission.s3_data_file_location}",
+        f"Details: {submission.s3_details_location}",
+        f"N Types: {load_data_response['ntypes']}",
+        f"Uniques: {load_data_response['unique']}",
```
willronchetti, line 73: Like mentioned today, this may be very long, you might want to just give a number?

willronchetti (approved):
Some small comments on recent work but still looks good

src/encoded/ingestion/loadxl_extensions.py
netsettler: Indeed. Having struggled with maintaining this in detail, I think we might want to consider testing merely that certain key lines are present, so having a list of regexps that are expected to appear at all, or in a given order. Probably if some of them are there, the others are.

src/encoded/tests/test_create_mapping.py
```diff
@@ -16,7 +16,7 @@
 pytestmark = [pytest.mark.setone, pytest.mark.working]
 
 
-@pytest.mark.parametrize('item_type', ITEM_INDEX_ORDER)
+@pytest.mark.parametrize('item_type', ITEM_INDEX_ORDER())
```
netsettler, line 19: This relates to my comment in other repos about turning this into a function. I really don't like this and would like to see either 
```
ITEM_INDEX_ORDER = ...call to other-named function that computes it...
```
or a direct call to that other-named function here.

The reason for all-uppercase is to say "this is constant" and the reason for `()` is to say "this is variable and must be computed", so this conflicted/confusing naming is what is stressing me as much as the change in semantics.

development.ini.template
```diff
@@ -7,7 +7,7 @@
 use = config:base.ini#app
 sqlalchemy.url = postgresql://postgres@localhost:5441/postgres?host=/tmp/snovault/pgdata
 blob_bucket = cgap-unit-testing-blobs
-metadata_bundles_bucket = cgap-unit-testing-metadata-bundles
+metadata_bundles_bucket = smaht-production-application-metadata-bundles
```
netsettler, line 10: This is a dev ini file but it's supposed to use the production metadata bundles bucket? That's not necessarily horrible, but I think the reason for wanting the analog of the cgap-unit-testing-metadata-bundles is so we can clean it from time to time and so people who lack production credentials can browse test data to debug it. @willronchetti, thoughts?

CHANGELOG.rst
```diff
@@ -26,6 +37,9 @@ Change Log
 0.0.8
 =====
 
+
+0.0.8
+=====
```
netsettler, line 42: I think this wrongly-duplicated text must have been an edit-o.

```suggestion
```

development.ini.template
willronchetti: I think an alternative bucket should be created yes similar to the devtest buckets

src/encoded/ingestion/loadxl_extensions.py
```diff
@@ -0,0 +1,75 @@
...
+        f"Successful ingestion summary:",
+        f"In File: {submission.data_file_name}",
+        f"S3 File: {submission.s3_data_file_location}",
+        f"Details: {submission.s3_details_location}",
+        f"N Types: {len(load_data_response['types'])}",
+        f"Created: {len(load_data_response['created'])}",
+        f"Updated: {len(load_data_response['updated'])}",
+        f"Skipped: {len(load_data_response['skipped'])}",
+        f"Checked: {len(load_data_response['validated'])}",
+        f"Errored: {len(load_data_response['errors'])}"
```
netsettler, line 74: Does it happen that we get errors in the middle even with stuff that passes the schema? That could be really messy.

src/encoded/ingestion/loadxl_extensions.py
```diff
@@ -0,0 +1,75 @@
...
+        overwrite=True,
+        itype=None,
+        from_json=True,
+        patch_only=False,
+        validate_only=validate_only)
+
+    return package_loadxl_response(loadxl_load_data_response)
+
+
+def summarize_load_data_into_database_response(load_data_response: Optional[dict],
```
netsettler, line 56: I don't like the name of this, which I don't find gives me a hint similar to the action or the doc string. Maybe (following from the doc string) `summary_of_load_data_results`? Or `summary_of_load_data_outcomes`?

src/encoded/ingestion/loadxl_extensions.py
```diff
@@ -0,0 +1,75 @@
...
+            # ASSUME each item in the loadxl response looks something like one of (string or bytes):
+            # POST: beefcafe-01ce-4e61-be5d-cd04401dff29
+            # PATCH: deadbabe-7b4f-4923-824b-d0864a689bb
+            # SKIP: feedbeef-eb17-4406-adb8-060ea2ae2180
+            # CHECK: cafebabe-eb17-4406-adb8-0eacafebabe
+            # ERROR: deadbeef-483e-4a08-96b9-3ce85ce8bf8c
+            # Note that SKIP means skip POST (create); it still may do PATCH (update), if overwrite.
+            if isinstance(item, bytes):
+                item = item.decode("ascii")
+            elif not isinstance(item, str):
```
netsettler, line 24: I think I would like a toplevel function like (untested):
```
def maybe_decode_ascii(str_or_bytes):
    if isinstance(str_or_bytes, bytes):
        return item.decode('ascii')
    return str_or_bytes
```
so you can here do:
```suggestion
            item = maybe_decode_ascii(item)
            if not isinstance(item, str):
```

or the function could be:
```
def maybe_decode_bytes(str_or_bytes, *, encoding='utf-8'):
    if isinstance(str_or_bytes, bytes):
        return item.decode(encoding)
    return str_or_bytes
```
so that the call can be
```
            item = maybe_decode_bytes(item, encoding='ascii')
            if not isinstance(item, str):
```

netsettler (approved):
Some of my comments probably merit consideration, but I don't think any are blocking.

pyproject.toml
```diff
@@ -67,6 +68,7 @@ pyasn1 = "0.1.9"
 PyBrowserID = "^0.14.0"
 pycparser = "2.14"
 PyJWT = "^2.6.0"
+#PyYAML = "^6.0.1"
 PyYAML = "5.3.1"
```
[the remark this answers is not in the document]
dmichaels-harvard: Removed this commented out cruft line; M1 needs 5.3.1 (or newer 6.x).

src/encoded/ingestion/loadxl_extensions.py
dmichaels-harvard: Done

src/encoded/ingestion/loadxl_extensions.py
dmichaels-harvard: Done

CHANGELOG.rst
dmichaels-harvard: Done

development.ini.template
dmichaels-harvard: Created bucket: smaht-unit-testing-metadata-bundles

src/encoded/ingestion/loadxl_extensions.py
dmichaels-harvard: Not entirely sure what mean - the only kinds of errors we could get for an item that passes the schema validation are errors during actual loading, which are collected/itemized in this module.

pyproject.toml
```diff
@@ -39,11 +39,12 @@ awscli = ">=1.25.36"
 boto3 = "^1.24.36"
 botocore = "^1.27.36"
 certifi = ">=2021.5.30"
-chardet = "3.0.4"
+chardet = "^5.0.0"
```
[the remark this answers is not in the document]
dmichaels-harvard: Release notes are here: https://github.com/chardet/chardet/releases
I didn't see anything too crazy, no mentions of breaking changes or backwards incompatibility (except Python now only supporting Python 3.7.+)

src/encoded/tests/test_create_mapping.py
dmichaels-harvard: I can change the name to not be upper-case, e.g. get_item_index_order(), but I needs to be a function because it is now coming in via the app_project mechanism (i.e. app_project().loadxl_order(), i.e. can change per downstream repo - smaht-portal, cgap-portal, fourfront).

dmichaels-harvard: Name change (to loadxl_order) made to this function.

PR #14, opened 2023-08-30
DRS + Unique/Name Key Fix

willronchetti:
- Allows use of DRS API with SMaHT Portal
- Fixes unique/name key resolution issues by porting unique_key implementation onto base Item class


src/encoded/tests/test_file_drs.py
```diff
@@ -0,0 +1,72 @@
...
+    validate_drs_conversion(drs_meta, file, uri=f'{DRS_PREFIX}/{file["uuid"]}')
+
+
+def test_fastq_file_drs_access(testapp, file):
+    """ Tests that access URLs are retrieved successfully """
+    drs_meta = testapp.get(file['@id'] + '@@drs').json
+    drs_object_uri = drs_meta['drs_id']
+    drs_object_download = testapp.get(f'/ga4gh/drs/v1/objects/{drs_object_uri}/access/').json
+    assert drs_object_download == {
+        'url': f'https://localhost:80/{drs_object_uri}/@@download'
```
netsettler, line 71: I feel like the various references to `https://localhost:80` (which incidentally is paradoxical since https uses 443, so I don't even know what that means) should be factored out into a variable or fixture so that it can be centrally updated (e.g., when you decide https: + port 80 was a bad idea.

netsettler (approved):
My comment is not blocking but much hoped-for. :)

PR #22, opened 2023-10-17
Bm homepage and more inserts

Bianca-Morris:
**Changelog:**
- Assorted edits to the styling of the homepage and navbars
- Moved a bunch of things around token/utility-wise
- Add Page & Static Section inserts for all Data and Documentation pages in the sitemap

**Demo:**

[Image from Gyazo](https://gyazo.com/fd6557426e2880e83d12078492711f36)


src/encoded/static/scss/encoded/modules/_tokens.scss
```diff
@@ -0,0 +1,43 @@
...
+
+$homepage-timeline-border-color: $seafoam-40;
+$homepage-timeline-bg-color: $seafoam-20;
+$homepage-timeline-text-color: $seafoam-60;
+
+$homepage-about-consort-btn-bg-color: rgba($pure-white, 0.75);
+$homepage-about-consort-btn-border-color: $cobalt-20; // based on mockup supposed to be #B1C3F8; this is closest
+$homepage-about-consort-btn-icon-color: $cobalt-30;
+$homepage-about-consort-text-color: $cobalt-70;
+$homepage-about-consort-border-radius: 4px;
```
Bianca-Morris, line 43: Should probably have added a comment about this; keeping these here for now, since I know they'll likely change pretty soon and it'll be easier to swap out and add colors when they're both in the same file. These will likely move to `base.scss`. I'll also probably add some tokens for things like font sizes, border-radius options (lg: 10px, sm: 4px, etc.)... anything we use often.

src/encoded/static/components/navigation/components/AccountNav.js
crfmc, line 1: Everything looks good to me. Just one question about how we should organize js files: I've been placing helper functions/child-components towards the top and the top-level component/parent at the bottom. The direction is not so important, but the order helps me a lot when reading react files. Could we settle on one direction? Or maybe some other logic we could stick to?

crfmc (approved):

src/encoded/static/components/navigation/components/AccountNav.js
Bianca-Morris: Hmm, yeah that's a good point. We can definitely try to do this, and adjust files as we work through them. Not sure it makes sense to do a whole-repo re-org at the moment, though, given how much there is to do.

PR #25, opened 2023-10-26
Cfm homepage figure

crfmc:
Associated in part with [Trello Ticket 818](https://trello.com/c/dafsbksc/818-smaht-home-page-implementation)
- Adds human figure and surrounding cards to homepage
- Imports SVG elements
- Temporary state variable [currentTier] to eventually track selected Tier
- Buttons for tier selection


src/encoded/static/components/viz/HomepageFigure/HumanFigure/Liver.js
```diff
@@ -0,0 +1,233 @@
+import React from 'react';
+import {
+    OverlayTrigger,
+    Popover,
+    PopoverTitle,
+    PopoverContent,
+} from 'react-bootstrap';
+
+const svg = (
```
Bianca-Morris, line 9: One thing about the way you've done the SVGs... depending on what we want to accomplish here it might be cleaner to just save the SVGs to files and import them. You can still do a fair amount of styling in that case (so, like if we want it to be a bit less opaque when not active, you should be able to do that in CSS still). Not sure what the default functionality of pointer events is like on SVGs with transparent/unfilled parts, but I know there's a way to use `pointer-events` to designate which parts you want to be clickable. So that hopefully wouldn't be an issue.

Bianca-Morris: If there's some other reason you did it this way, though, let me know. I'd be curious if it has some benefit I don't know about.

Bianca-Morris (approved):

PR #29, opened 2023-11-06
Submittable Data Model

drio18:
Here, we add a first cut for "submittable" items, i.e. items we expect external users to submit to us.

Main components of this PR include:

- Schemas and types for submittable items
- Permissions fixes and updates
- Tests for the above + more robust coverage for various fixes and updates

Some of the primary ideas for the data model are:

- Items to be submitted generally have a dependency chain via property requirements. For example, to submit files, one must also submit or have had submitted a file set.
- All items to be submitted must have a `submitted_id` which serves as a unique identifier for the submitters
- The majority of previously existing items (a.k.a. the "core" items) cannot be submitted by non-admins.
- For now, there is no central or collecting item type, like a Case on CGAP or an Experiment on 4DN.

For testing, changes of interest include:

- All items to be represented in workbook inserts, to serve as source for majority of tests instead of pytest fixtures.
- Added broader, more generic testing to cover create permissions, schema expectations, unique keys, etc. rather than repeating similar tests for individual item types.
- Deleted many existing tests due to changes above.


willronchetti (approved):
Overall this is exceptional work. There is very little to scrutinize throughout this, the few comments I did make have to do with some specific design decisions related to files that just maybe need further clarification on the thinking. Tests look great. I think this is definitely a good place to start with things.

aschroed:
It looks like you should update the image schema to remove aliases and add submitter ID as it is referenced in histology (and maybe elsewhere?)

src/encoded/types/cell_sample.py
```diff
@@ -0,0 +1,17 @@
...
+
+@collection(
+    name="cell-samples",
+    unique_key="submitted_id",
+    properties={
+        "title": "Cell Samples",
+        "description": "Samples consisting of isolated cells",
+    },
+)
+class CellSample(Sample):
```
aschroed, line 14: Does this 'inheritance' work as expected?  I ask because last I knew non-abstract items could only inherit from abstract items.  Great if this is no longer the case for SMaHT but wanted to make sure.

aschroed (approved):
This looks really good.  I may add additional comments here and there but they would be mostly to clarify things and don't see anything blocking.  

It would be nice to view a schema wire diagram to be able to visualize the linking - @willronchetti may have a quick way to generate it?  But if not not a problem and not worth taking the time.  Everything I've looked at in that respect looks good.

src/encoded/types/cell_sample.py
drio18: I believe so, as I think we have non-abstract inheritance in CGAP with WorkflowRun and WorkflowRunAwsem. Sample is an abstract collection here though.

PR #31, opened 2023-11-15
Fixes for Wolf

willronchetti:
- Fixes various issues detected in bioinformatics testing


src/encoded/types/file.py
```diff
@@ -60,3 +92,134 @@ def upload_credentials(self) -> Union[str, None]:
...
+    """validator to make sure that the values in the
+    produced_from field are valid file identifiers"""
+    # skip validator if not file processed
+    if context.type_info.item_type != 'output_file':
+        return
+    data = request.json
+    if 'produced_from' not in data:
+        return
+    files_ok = True
+    files2chk = data['produced_from']
```
dmichaels-harvard, line 167: Just sort of style comment: I usually do this kind of thing just so as not to look it (produced_from) up in data twice: 
`if not (produced_from := data.get("produced_from")):
    return ;
use__produced_from__here  ...`

dmichaels-harvard (approved):
LGTM

src/encoded/types/output_file.py
```diff
@@ -14,3 +14,10 @@ class OutputFile(File):
     item_type = "output_file"
     schema = load_schema("encoded:schemas/output_file.json")
     embedded_list = []
+
+    # processed files don't want md5 as unique key
+    def unique_keys(self, properties):
+        keys = super(OutputFile, self).unique_keys(properties)
+        if keys.get('alias'):
```
drio18, line 21: Should be `aliases` here? Also, rather than do this here, perhaps we should update tibanna not to patch an md5 alias instead of handling here if that's the source of the issue.

src/encoded/types/file.py
```diff
@@ -60,3 +92,134 @@ def upload_credentials(self) -> Union[str, None]:
...
+            files_ok = False
+            request.errors.add('body', 'File: invalid produced_from id', "'%s' not found" % f)
+            # bad_files.append(f)
+        else:
+            if not fid:
+                files_ok = False
+                request.errors.add('body', 'File: invalid produced_from id', "'%s' not found" % f)
+
+    if files_ok:
+        request.validated.update({})
```
drio18, line 181: Since we're trying to have shared code via encoded-core, consider refactoring in encoded-core to make these validators shared by passing in the item type name instead of copy-pasting with minor change here. 

Can also consider cleaning some of this up and/or removing parts that don't really make sense for this portal (like checking `produced_from` ?).

src/encoded/types/file.py
```diff
@@ -60,3 +92,134 @@ def upload_credentials(self) -> Union[str, None]:
...
+        conn = request.registry['connection']
+        res = conn.get_by_json('md5sum', data['md5sum'], 'output_file')
+        if res is not None:
+            # md5 already exists
+            found = res.properties['accession']
+            request.errors.add('body', 'File: non-unique md5sum', 'md5sum %s already exists for accession %s' %
+                               (data['md5sum'], found))
+
+
+def validate_processed_file_produced_from_field(context, request):
```
drio18, line 157: Do we need this validator? We don't have any schema field for `produced_from` for files.

src/encoded/types/file.py
```diff
@@ -1,15 +1,44 @@
...
 )
-from pyramid.request import Request
-from snovault import abstract_collection, calculated_property, load_schema
+from encoded_core.file_views import (
+    validate_file_filename,
+    validate_extra_file_format,
+    drs as CoreDRS,
+    download as CoreDownload,
+    post_upload as CorePostUpload,
+    get_upload as CoreGetUpload
```
drio18, line 16: Consider adding tests for these to ensure working as desired and aren't unintentionally broken in future.

drio18 (commented):
Mostly non-blocking suggestions and questions, but I think some quick tests of function for the file views/validators are a good idea and a property name reference needs to be changed.

src/encoded/types/file.py
```diff
@@ -37,6 +66,9 @@ class File(SMAHTItem, CoreFile):
 
     SHOW_UPLOAD_CREDENTIALS_STATUSES = ("in review",)
 
+    class Collection(SMAHTItem.Collection):
```
drio18, line 69: Also, I think we should update line 67 above to appropriate statuses included in this PR, namely `uploading`, `to be uploaded by workflow`, and `upload failed` (not entirely sure about the latter, but these are the 3 we use in encoded-core). There's a corresponding test that will need to be updated and fixed if done here.

src/encoded/types/file.py
```diff
@@ -37,6 +66,9 @@ class File(SMAHTItem, CoreFile):
...
+        pass
+
```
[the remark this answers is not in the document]
willronchetti: An override is needed in this type definition in order to bind a view config to this Collection and not the one in base.py, otherwise the below routes for file with custom validators throw a configurator error

src/encoded/types/output_file.py
willronchetti: The issue is generally for files we want md5 to be unique key, except in the case of processed files because we want to be able to rerun pipelines. This code is the same as in CGAP and I believe should work, the unit test seems to think so

src/encoded/types/file.py
willronchetti: We can't refactor to use in encoded-core at this time because the original definition hard codes the type ie: in encoded-core, it has the CGAP implementation and hard codes processed_file, hence why this didn't work originally. I can refactor this eventually though.

src/encoded/types/file.py
willronchetti: Must be a holdover from 4DN, CGAP has it and doesn't seem to affect... I can remove though

src/encoded/types/file.py
willronchetti: This will come in an upcoming PR very soon, for now I'd like to isolate the status related changes

src/encoded/types/file.py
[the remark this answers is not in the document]
drio18: Ah, makes sense, thanks for clarifying.

src/encoded/types/output_file.py
drio18: Do we need md5 to be a unique key for looking up the file? Maybe we should update `File.unique_keys` not to add that in and just do this via the schema. I don't know; this works, just kind of hacky and unclear if the complexity is necessary here.

src/encoded/types/file.py
drio18: Yeah, I thought it would be cleaner to refactor in encoded-core to pass the type so could be re-used easily in smaht.

PR #32, opened 2023-11-16
Permissions Take 2

willronchetti:
- Prune status as previously discussed
- Add custom validation for consortia/submission centers
- Expand unit tests


src/encoded/docs/permissions.rst
```diff
@@ -0,0 +1,67 @@
+========================
+smaht-portal permissions
+========================
+
+Permissions in the smaht-portal rely on 3 mechanisms:
+    * Entity tagging with consortia, submission center
+    * The "status" field
```
dmichaels-harvard, line 7: Very minor: "submission centers" plural to match "consortia".

src/encoded/docs/permissions.rst
```diff
@@ -0,0 +1,67 @@
...
+    * The admin group
+
+For the most part, those internal to the DAC are admins and can/view edit
```
dmichaels-harvard, line 10: Maybe define DAC parenthetically the first time used.

src/encoded/types/meta_workflow.py
```diff
@@ -1,11 +1,12 @@
 from snovault import collection, load_schema
 from encoded_core.types.meta_workflow import MetaWorkflow as CoreMetaWorkflow
-
+from .base import SUBMISSION_CENTER_MEMBER_CREATE_ACL
 from .base import Item as SMAHTItem
 
 
 @collection(
     name='meta-workflows',
+    acl=SUBMISSION_CENTER_MEMBER_CREATE_ACL,
```
alexander-veit, line 9: Does that mean that Submission Centers can create their own MetaWorkflows? Do we want that?

alexander-veit (approved):
Looks good to me.

src/encoded/tests/datafixtures.py
```diff
@@ -118,7 +128,20 @@ def smaht_admin(testapp):
 
 
 @pytest.fixture
-def smaht_gcc_user(testapp, test_submission_center):
+def blank_user(testapp):
+    item = {
+        'first_name': 'Unaffiliated',
+        'last_name': 'User',
+        'email': 'unaffiliated@example.org',
+        'status': 'current'
```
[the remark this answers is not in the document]
willronchetti: User is an exception to the status refactor in that it only uses the special statuses: current, deleted, inactive and revoked. Probably could cut this down additionally to current and deleted though.

src/encoded/types/meta_workflow.py
willronchetti: I had thought about this... Since this doesn't involve actual instantiation of workflows it might be useful to allow centers to do this if they want to contribute back pipelines. Could always be added later though.

alexander-veit: Right, they can't create MetaWorkflowRuns. I think that's fine then.

drio18 (approved):
Looks good. 

Going through all of this makes me wonder whether we should add the user's consortia as currently implemented upon item creation. This is more of a process matter, in the sense of who should control these entity tags and when/how they should be added. Seems like something we need to handle directly as the DAC rather than relying on incoming defaults.

src/encoded/types/higlass_view_config.py
```diff
@@ -1,11 +1,12 @@
 from snovault import collection, load_schema
 from encoded_core.types.higlass_view_config import HiglassViewConfig as CoreHiglassViewConfig
-
+from .acl import ALLOW_CONSORTIUM_CREATE_ACL
 from .base import Item as SMAHTItem
 
 
 @collection(
     name='higlass-view-configs',
+    acl=ALLOW_CONSORTIUM_CREATE_ACL,
```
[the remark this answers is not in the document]
willronchetti: Intended to allow all consortia users to create new higlass views, future proofing and probably not necessary now

src/encoded/docs/permissions.rst
```diff
@@ -0,0 +1,67 @@
...
+Released is intended for data that should be viewable by all consortia members.
+
+In review is the default status for uploaded data and is only viewable and
+editable by the submitter and the DAC.
+
+Obsolete is a form of released where data has either been superseded in some
+way or is only up for historical purposes.
+
+Deleted is a special status for data that should not be viewable by anyone
+except the DAC and is kept only for historical purposes.
```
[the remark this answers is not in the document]
willronchetti: addressed

PR #33, opened 2023-11-16
More schema updates

drio18:


src/encoded/schemas/static_section.json
```diff
@@ -102,14 +100,6 @@
...
     ],
     "properties": {
-        "category": {
-            "items": {
-                "type": "string",
-                "enum": [
-                    "Page Section"
-                ]
-            }
-        },
```
willronchetti, line 112: We should double check with UI this is ok, though it does seem useless to me

src/encoded/schemas/workflow.json
```diff
@@ -222,17 +226,7 @@
...
-                            "c5.4xlarge",
-                            "m5a.2xlarge",
-                            "r4.4xlarge",
-                            "r5.4xlarge",
-                            "r5a.4xlarge",
-                            "r5b.4xlarge",
-                            "r5ad.4xlarge",
-                            "r6i.4xlarge"
-                        ]
+                        "pattern": "^[a-z][a-z0-9-]+[.][0-9]*[a-z-*]+$"
```
willronchetti, line 229: I get the idea here but we may want a custom validator for this as if the intention is to validate for an expanded set of instances it will additionally validate instances that cannot often be used if at all (ie: GPU instances, Graviton etc).

src/encoded/tests/test_schema_document.py
```diff
@@ -0,0 +1,25 @@
...
+        "consortia": [test_consortium["uuid"]],
+        "submission_centers": [test_submission_center["uuid"]],
+        "description": "Some document",
+    }
+    return post_item(testapp, item, "Document")
+
+
+def test_document_post(document: Dict[str, Any]) -> None:
+    """Ensure Document properties POST."""
+    pass
```
willronchetti, line 25: You might want to assert something basic in these tests ie: description matches or existence of calc prop. Same deal for other tests of this nature.

src/encoded/tests/test_schema_quality_metric.py
```diff
@@ -0,0 +1,28 @@
...
+    test_consortium: Dict[str, Any],
+    test_submission_center: Dict[str, Any],
+) -> Dict[str, Any]:
+    item = {
+        "consortia": [test_consortium["uuid"]],
+        "submission_centers": [test_submission_center["uuid"]],
+        "category": "Testing",
+        "qc_values": [{"key": "some_qc_metric", "value": "22"}],
+        "overall_quality_status": "Pass",
+        "url": "https://foo.bar",
```
willronchetti, line 21: We might want validate/check that this is an S3 URL

src/encoded/tests/utils.py
```diff
@@ -21,33 +23,114 @@ def post_item(
...
     return response.json
 
 
-def get_formatted_resource_path(resource_path: str, add_on: Optional[str] = None) -> str:
-    resource_path_with_add_on = get_resource_path_with_add_on(resource_path, add_on)
+def get_frame_add_on(frame: Union[str, None]) -> str:
+    """Format frame parameter, if provided."""
+    if frame:
+        return f"frame={frame}"
+    return ""
```
willronchetti, line 58: Probably not useful but you might want to validate this is one of `page`, `embedded`, `object` or `raw`

src/encoded/tests/utils.py
```diff
@@ -21,33 +23,114 @@ def post_item(
...
+        status: Optional[int] = None,
 ) -> Dict[str, Any]:
-    resource_path = get_formatted_resource_path(identifier, add_on=f"frame={frame}")
-    response = testapp.get(resource_path)
+    """Get item view with given frame, following redirects."""
+    add_on = get_frame_add_on(frame)
+    resource_path = get_formatted_resource_path(
+        identifier, collection=collection, add_on=add_on
+    )
+    response = testapp.get(resource_path, status=status)
```
willronchetti, line 48: Consider adding `datastore` argument as well for combined ES/Postgres tests that want to request `datastore=database`

willronchetti (approved):
Looks good, no blocking comments, some small notes

src/encoded/schemas/workflow.json
drio18: Agreed. Bioinformatics seemed to prefer the regex to the hard list, since adding a new instance would require an update and deployment, but there's more potential for using an unsupported instance or one with a typo with the regex. On CGAP, we never verified anything and did okay; will revisit as needed.

PR #35, opened 2023-11-20
SMaHT ingestion work

dmichaels-harvard:
Note that this branch also has merged in changes from Doug's ``drr_submitted_items`` branch.


src/encoded/ingestion/structured_data.py
```diff
@@ -0,0 +1,606 @@
...
+            column_name_component = column_name_components[0]
+            array_name, array_index = _StructuredRowData._get_array_info(column_name_component)
+            name = array_name if array_name else column_name_component
+            if len(column_name_components) > 1:
+                if not isinstance(row[name], dict) and not isinstance(row[name], list):
+                    row[name] = {}
+                setv(row[name], column_name_components[1:], parent_array_index=array_index)
+                return
+
+            nonlocal column_name, value, schema, loc
```
willronchetti, line 172: Neat usage, though I think you only need `nonlocal column_name` since this is the only ref outside the nested function?

src/encoded/ingestion/structured_data.py
```diff
@@ -0,0 +1,606 @@
...
+    @staticmethod
+    def _get_array_info(name: str) -> Tuple[Optional[str], Optional[int]]:
+        if (array_indicator_position := name.rfind(ARRAY_NAME_SUFFIX_CHAR)) > 0:
+            array_index = name[array_indicator_position + 1:] if array_indicator_position < len(name) - 1 else -1
+            if (array_index := to_integer(array_index)) is not None:
+                return name[0:array_indicator_position], array_index
+        return None, None
+
+
+class Schema:
```
willronchetti, line 218: Is this class at all necessary? What does this do that `load_schema` doesn't?

src/encoded/tests/test_structured_data.py
```diff
@@ -0,0 +1,592 @@
...
+        }
+    )
+
+def test_parse_structured_data_1():
+    _test_parse_structured_data(sheet_utils_also = False,
+        as_file_name = "some_test.csv",
+        rows = [
+            r"uuid,status,principals.view,principals.edit,extensions#,num,i,arr",
+            r"some-uuid-a,public,pav-a,pae-a,alfa|bravo|charlie,123.4,617,hotel",
+            r"some-uuid-b,public,pav-b,pae-b,delta|echo|foxtrot|golf,987,781,indigo\|juliet|kilo"
```
willronchetti, line 57: This is mostly a stylistic comment, but seeing how many of these there are I do think you should consider using `@pytest.mark.parametrize` to do most of these no? It will do this sort of numbering thing for you. Because of the data size it might be less clean in practice but I think overall may be easier to see what's going on in the big picture.

willronchetti (approved):
I think some of these merit addressing before merging but they are not strictly speaking blocking

src/encoded/ingestion/structured_data.py
dmichaels-harvard: The main point of this class is to create/contain the stripped-down "flattened" version of the schema for quick lookup of types to convert values in the data appropriate to the schema (via map_value method); and also to validate data using the loaded schema (via validate method). But yes the part that gets/loads the schema uses the ff_utils.get_schema method, which ultimately does end up using/calling (snovault) load_schema.

src/encoded/ingestion/structured_data.py
dmichaels-harvard: No I don't think so; value, schema, and loc are used as well (unless I'm misunderstanding your question).

src/encoded/commands/parse_structured_data.py
```diff
@@ -0,0 +1,263 @@
...
+            print(json.dumps(schema, indent=4, default=str))
+        else:
+            print(f">>> No schema found for type: {schema_name}")
+
+
+def _create_portal_for_local_testing(ini_file: Optional[str] = None, schemas: Optional[List[dict]] = None) -> Portal:
+    if isinstance(ini_file, str):
+        return Portal(create_testapp(ini_file), schemas=schemas)
+    minimal_ini_for_local_testing = "\n".join([
+        "[app:app]\nuse = egg:encoded\nfile_upload_bucket = dummy",
```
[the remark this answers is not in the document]
dmichaels-harvard: I kinda like it being self-contained without external (file) dependency, unless you feel strongly about it.

src/encoded/commands/parse_structured_data.py
```diff
@@ -0,0 +1,263 @@
...
+            SchemaManager.get_schema = lambda name, portal_env, portal_vapp: {}
+
+    # Manually override implementation specifics for our default handling of refs (linkTo),
+    # which is to catch/report any ref errors; use --norefs to not do ref checking at all;
+    # and use --default-refs to throw exceptions for ref errors (as normal outside of this script).
+    refs = override_ref_handling(args) if args.norefs or not args.default_refs or args.show_refs else None
+
+    if args.verbose:
+        if args.sheet_utils:
+            print(f">>> Using sheet_utils rather than the newer structured_data ...")
```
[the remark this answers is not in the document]
dmichaels-harvard: Yes, thanks.

src/encoded/tests/test_structured_data.py
dmichaels-harvard: Ah, I didn't realize that benefit, of numbering output - I will check it out, thanks.

PR #36, opened 2023-11-20
Small Perm Fix

willronchetti:
- Fix statuses in inserts
- Add INGESTER remoteuser to validator bypass

### DIFF
--- a/src/encoded/types/base.py
+++ b/src/encoded/types/base.py
@@ -254,7 +254,7 @@ def validate_user_submission_consistency(context, request):
         part of
     """
     user = request.environ.get('REMOTE_USER')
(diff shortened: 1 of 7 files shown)


src/encoded/types/base.py
```diff
@@ -254,7 +254,7 @@ def validate_user_submission_consistency(context, request):
         part of
     """
     user = request.environ.get('REMOTE_USER')
-    if not user or user in ['TEST', 'INDEXER', 'EMBED', 'TEST_SUBMITTER']:
+    if not user or user in ['TEST', 'INDEXER', 'EMBED', 'TEST_SUBMITTER', 'INGESTER']:
```
dmichaels-harvard, line 257: That's it. Thanks!

dmichaels-harvard (approved):

PR #37, opened 2023-11-27
Repair custom_pf_fields

willronchetti:
- Remove a ``$merge`` that is causing propagating of additional fields causing tibanna to break


willronchetti:
Closing in favor of updates from Doug on another branch

PR #39, opened 2023-12-08
Cfm homepage updates

crfmc:
Style (+ responsive) updates to:
- Navigation bar
- Timeline
- Homepage figure
- User page

### DIFF
--- a/src/encoded/static/components/app.js
+++ b/src/encoded/static/components/app.js
@@ -265,18 +265,6 @@ export default class App extends React.PureComponent {
 
         console.info('BROWSER', browserInfo);
 
-        Alerts.queue({
-            style: 'danger',
-            message: (
-                <>
-                    <b>Attention Users:</b> This is an unofficial release of the
-                    SMaHT Data Portal, made available only to provide the access
-                    keys for metadata submission and testing. SMaHT data and
-                    additional information is coming soon.
-                </>
-            ),
-        });
-
         if (
             browserInfo &&
             typeof browserInfo.name === 'string' &&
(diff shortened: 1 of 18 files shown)


Bianca-Morris (commented):
In addition to Shannon's requested changes, I have a few small things:
     - Looks like the changes you made to the homepage alert container have resulted in some spacing issues among alerts that we already have. [These alerts are coming from SPC](https://github.com/4dn-dcic/shared-portal-components/blob/b5afc27bd7548fbd40047f2c0ee04164d3a1a4dc/src/components/ui/Alerts.js#L190-L220), and as such make use of the `title` prop. I don't mind you adding custom JSX to the `message` of the new alert to get your custom subtitle on the same line if that's what we want for these specific ones, but we should otherwise maintain the previous hierarchy of those titles. With alerts of larger size/more content, the title being inline will look less appropriate than it does with one-or-two liners (e.g. maintenance alerts, which can run pretty long). In the future we may want to add a subtitle prop to SPC, as well, if we find we want to do what you did here often.
     [Image from Gyazo](https://gyazo.com/33884a016648a1c48110c564d8d05a15)
    - The `width` or `height` attribute on the logo image served a purpose. It was there less to style the logo and more to prevent a flash of the full-size logo on page load before styling has fully loaded in. Alternatively, you could re-size the image to be the max size it is on load, and that should also resolve it. You may not have noticed this because it only happens when you cache clear and load everything for the first time (and is also most obvious on slow connections).
    - Also noticing that odd wrapping is sometimes happening on the footer. Not a huge priority. Especially since we have a variety of other mobile fixes to do for the top nav, in general in the near future and these things can be bundled into them.
    [Image from Gyazo](https://gyazo.com/68c3b4fe2729ffd6154a609b2858c982)
   - This requires no changes for now, but I will say, things like `AssaysCard` and `CLTCard` probably don't need their own folders. As a general rule, if there's only going to be one file in a folder, don't create it unless it really doesn't make sense to go elsewhere. It doesn't seem like either of these things are going to be expanded much further (if I'm wrong on this let me know), so having a separate folder for each seems like overkill. I'd either name one folder `Cards`, and have both in it. Or just have the two files live in `/HomepageFigure`. (Probably the latter would be preferred)
   - Did we decide against adding additional Y padding to the figure + timeline? Feels a little cramped on top to me.

Bianca-Morris (approved):

PR #41, opened 2023-12-12
Fix to test_structured_data.py (WRT testapp).

dmichaels-harvard:
Had previously removed so the whole file appears here. Key thing is using the "testapp" fixture.


pyproject.toml
```diff
@@ -44,9 +44,9 @@ certifi = ">=2021.5.30"
 chardet = "^5.0.0"
 codeguru-profiler-agent = "^1.2.4"
 colorama = "0.3.3"
-dcicsnovault = "^11.6.0"
-dcicutils = "^8.5.0"
-encoded-core = "^0.4.0"
+dcicsnovault = "^11.6.0.1b3"
+dcicutils = "^8.5.0.1b2"
+encoded-core = "^0.4.0.1b3"
```
willronchetti, line 49: Should get real versions when possible

willronchetti (approved):
LGTM

PR #44, opened 2023-12-13
Pre-release updates 

crfmc:
Changes made for the pre-release
- feat: update human figure
- style: add arrows in to timeline headers
- style: make timeline markers circular


Bianca-Morris (approved):
Looking really good!

I did have one question regarding the arrow thing. I thought we had decided at one point to have the arrows not be visible for users that were not logged in, since they won't have access to all data sets? If that decision was overridden at some point, disregard this.

PR #47, opened 2023-12-14
Fix admin affilitation validation

drio18:
Fix validator for affiliations on item creation/edits to remove admins from the restriction.


willronchetti (approved):
Good catch!

PR #48, opened 2023-12-14
More QC value types

drio18:
Allow any non-object JSON type for QC values instead of just strings.


alexander-veit (approved):
Looks good.

PR #51, opened 2023-12-21
QC Ruleset Validation

drio18:
Here, we add support for QC rulesets as pipeline inputs and validate them appropriately. The validation is performed via `if/then` [logic in the schema](https://json-schema.org/understanding-json-schema/reference/conditionals#ifthenelse), which I believe is new for our projects but works as expected per tests.

Note: Discussed use of separate item type for QC rulesets with bioinformatics rather than the nested object use here; decision was to move forward with this implementation and consider allowing separate items in the future. Main benefits of separate items would be easier editing of the properties and improved sharing of the rulesets.


src/encoded/schemas/mixins.json
```diff
@@ -249,6 +249,120 @@
...
+                                                "type": "string"
+                                            },
+                                            "metric": {
+                                                "title": "Metric",
+                                                "type": "string"
+                                            },
+                                            "operator": {
+                                                "title": "Operator",
+                                                "type": "string",
+                                                "enum": [
```
alexander-veit, line 317: The currently supported operators are `==`, `<`, `>`, `<=`, `>=`, `!=`

src/encoded/schemas/mixins.json
```diff
@@ -249,6 +249,120 @@
...
+                                            "use_as_qc_flag": {
+                                                "title": "Use as QC Flag",
+                                                "type": "boolean"
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                        },
+                        "value_type": {
```
alexander-veit, line 358: This is not really clear to me, but I assume it is correct?

alexander-veit:
Do we need to add this verification logic to the MetaWorkflowRuns as well? We will probably need to overwrite the rulesets for different runs, i.e. it will be on the run item.

src/encoded/schemas/mixins.json
alexander-veit: Ok... got it now.

alexander-veit (approved):

PR #56, opened 2024-01-08
Submittable item API for SMaHT portal

aschroed:
Implementation of submittable item API for SMaHT portal.

@drio18 would like you to look at the values used in the project/schema_views.py
using 'submitted_id' as the key prop to identify a schema/item as submittable
Note: I've included submission_centers and consortia in excluded_props thinking submittr would add these based on who was running the tool but not sure if that is true?
Also filtering out calculatedProperty: True and permissions: restricted_fields

Wasn't sure if maybe allowing the project values to SchemaViews to be provided as parameters was a good idea and how difficult that might be to implement if so?

I added a couple of 'functional' tests.  Not sure what more should be added and if these should really be mocked up using the structures that @willronchetti set up or if this is sufficient?


src/encoded/tests/test_schemas.py
```diff
@@ -220,3 +219,48 @@ def test_unique_keys_are_identifying_properties(testapp: TestApp) -> None:
...
+def test_submittables(testapp, registry):
+    test_uri = '/submission-schemas/'
+    expected_items = [
+        'AlignedReads', 'Analyte', 'AnalytePreparation', 'CellCulture',
+        'CellCultureMixture', 'CellCultureSample', 'CellLine', 'CellSample',
+        'DeathCircumstances', 'Demographic', 'Diagnosis', 'Donor', 'Exposure',
+        'FileSet', 'Histology', 'Library', 'LibraryPreparation', 'MedicalHistory',
+        'MolecularTest', 'PreparationKit', 'SamplePreparation', 'Sequencing',
+        'Software', 'Therapeutic', 'Tissue', 'TissueCollection', 'TissueSample',
+        'Treatment', 'UnalignedReads', 'VariantCalls']
```
drio18, line 262: Consider replacing with [this function](https://github.com/smaht-dac/smaht-portal/blob/main/src/encoded/tests/utils.py#L254-L264) to make easier to maintain.

src/encoded/project/schema_views.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def get_submittable_item_names(self):
+        return []
+
+    def get_prop_for_submittable_items(self):
+        return 'submitted_id'
+
+    def get_properties_for_exclusion(self):
+        return ['date_created', 'submitted_by', 'uuid', 'schema_version',
+                'submission_centers', 'consortia', 'principals_allowed']
```
drio18, line 14: I think these are good; some of them are redundant with the exclusion attributes, but that's fine.

src/encoded/tests/test_schemas.py
```diff
@@ -220,3 +219,48 @@ def test_unique_keys_are_identifying_properties(testapp: TestApp) -> None:
 def has_identifying_property(schema: Dict[str, Any], property_name: str) -> bool:
     """Return True if schema has property_name property."""
     return property_name in schema_utils.get_identifying_properties(schema)
+
+
+def test_submittable(testapp, registry):
```
drio18, line 224: I think this test is a bit redundant with the ones in snovault and will be a minor burden to maintain when things change on the given item type. Consider just grabbing one of the item types from the items with `submitted_id` and checking that the `/submission-schemas/{item_type}.json` endpoint works and looks grossly correct (e.g. lacks all excluded properties, lacks properties with given attributes, etc.) without exact verification of exact properties and property counts.

drio18 (commented):
Dropping a few comments for consideration. Will deploy locally to take a final look at how things are working in practice before approving.

src/encoded/tests/test_schemas.py
drio18: May also be good to check that a non-submittable item fails this endpoint as expected.

src/encoded/tests/test_schemas.py
aschroed: done

src/encoded/tests/test_schemas.py
aschroed: Added test for expected 404 on non-submittable item.

Regarding the redundancy of the test - I think it's not a bad thing.  However, I take your point on it needing updates if things change.  That was part of why I asked about possible mocking and using a test schema that would not change - or I could certainly simplify as you suggest.  Could go either way.

drio18: Right, sorry, I missed that on my first time through. My vote is for the second option (using a test schema combined with some snovault mocking) to ensure the project defaults and the endpoint are working as intended.

src/encoded/__init__.py
```diff
@@ -109,6 +109,7 @@ def static_resources(config):
     mimetypes.init([pkg_resources.resource_filename('encoded', 'static/mime.types')])
     config.add_static_view('static', 'static', cache_max_age=CGAP_STATIC_MAX_AGE)
     config.add_static_view('profiles', 'schemas', cache_max_age=CGAP_STATIC_MAX_AGE)
+    config.add_static_view('can_submit', 'submittables', cache_max_age=CGAP_STATIC_MAX_AGE)
```
willronchetti, line 112: While we're here, we may consider renaming this

src/encoded/project/schema_views.py
willronchetti: Status and accession as well perhaps? Though these are covered by the exclusion attributes as well, it may be more clear in error messaging if explicitly included here.

src/encoded/tests/test_schemas.py
```diff
@@ -220,3 +221,47 @@ def test_unique_keys_are_identifying_properties(testapp: TestApp) -> None:
...
+def test_submittables(testapp):
+    test_uri = '/submission-schemas/'
+    expected_items = [to_camel_case(i) for i in get_items_with_submitted_id(testapp)]
+    res = testapp.get(test_uri, status=200).json
+    assert len(expected_items) == len(res)
+    for item in res.keys():
+        assert item in expected_items
+
+
+def test_not_submittable(testapp):
```
willronchetti, line 264: You might consider using `pytest.mark.parametrize` to quickly test all non-submittable types here

willronchetti (approved):

drio18 (approved):
Looks good when run locally, and thank you for moving to a test schema.

PR #57, opened 2024-01-09
Submitted ID Validation

drio18:
Here, we add validation for the `submitted_id` property to enforce certain standards for all submittable item types. Essentially, this standardizes and enforces concepts used more loosely for aliases in 4DN and CGAP. In particular, validation consists of:

- SubmissionCenter "code" will be included and validated against the SubmissionCenters on the item
- Item type included in `submitted_id` to prevent namespacing issues (validated per item-specific regex)
- Identifier of submitter's choosing validated per item-specific regex

To do so, a new item type, `SubmittedItem` is created and used as a parent class for all types to be submitted.

Tests are added to ensure consistency of `submitted_id` patterns and use for all `SubmittedItems`.

TODO: Will need to patch existing items' `submitted_ids` prior to deploying to ensure consistency, as well as deploy a branch with the new required field on SubmissionCenters that has the field but not as a requirement to add those.


src/encoded/tests/test_structured_data.py
```diff
@@ -19,6 +17,23 @@
 SAME_AS_EXPECTED_REFS = {}
 SAME_AS_NOREFS = {}
 
+# Ideally this flag would be False, i.e. we would like to use the actual/real (live)
+# schemas that are defined in (the schemas directory of) this repo for testing; but at
+# least currently (2024-01-10) these are undergoing a lot of change and leading to frequent
+# and annoying test breakage; so setting this to True will cause these tests to use a dump of
+# the schemas which were previously saved into a static file (data/test-files/schemas_dump.json).
+USE_SAVED_SCHEMAS_RATHER_THAN_LIVE_SCHEMAS = True
```
willronchetti, line 25: I think this is fine but could we at least read the file from dcicutils (it is in one of David's PRs as well) so we don't have this huge dump file replicated across repos? Referencing the exact same file could come with advantages as well so we ensure there is no divergence while still actively developing.

src/encoded/types/submitted_item.py
```diff
@@ -0,0 +1,284 @@
...
+    validate_item_content_put,
+    validate_item_content_patch,
+)
+
+from .base import Item, SMAHTCollection, collection_add, item_edit
+from .utils import get_item, get_properties
+
+
+"""
+SubmittedItems are the item types we expect to receive from external
```
willronchetti, line 20: Very useful info and should go in RST/portal documentation at some point

willronchetti (approved):
These changes generally look great and make sense, just a few small comments

src/encoded/commands/update_submitted_id_pattern.py
```diff
@@ -0,0 +1,322 @@
...
+    schema_path: Path, schema_contents: Dict[str, Any]
+) -> None:
+    """Write schema content to path."""
+    with schema_path.open("w") as file_handle:
+        json.dump(schema_contents, file_handle, indent=4)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
+        description="Update `submitted_id` pattern in all schemas"
```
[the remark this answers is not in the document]
aschroed: I was also curious about this.  Will this not potentially lead to a mass of validation errors - I don't know the status of upgraders in SMaHT but depending on the use case there would need to be some coupling there if changes are made after items have been submitted?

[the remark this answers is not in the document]
drio18: Mostly, I decided to automate it because I didn't want to do the manual work, and if things change then the update will be simpler in the future.

Changes could cause validation errors, but that's not necessarily the case; depends on if the changes are backwards compatible or not.

src/encoded/tests/test_structured_data.py
drio18: David added these changes, so will discuss with him.

PR #63, opened 2024-01-22
Benchmarking schema updates

drio18:
This PR makes several data model changes required for the benchmarking data release, including:

- Two new item types to segment properties more cleanly and capture the "code" information used in the file naming scheme as metadata
- New `code` property on many item types to capture as above
- A few other new properties required for released file search
- File columns, facets, and embedded fields for file search

Some changes (updated enums) do require upgraders to properly clean up properly without breaking changes, but front-end is dependent on these updates going in for their work, so will create a separate PR for this to follow shortly.


aschroed (commented):
A portion of the parameterized test_structured_data tests are failing.  Looks like the mocking of dcicutils.structured_data.Schema._map_function_ref is not working as expected?

willronchetti (approved):
I have no issues with this but can't repro the issue Andy pointed out, GA looks clean so once main merged in should check again before merge

aschroed (approved):
If the test issues I saw were not reproducible and therefore likely due to my local setup then no worries.  Note that it may be worth updating encoded-core version in pyproject.toml though did not make a difference with the test issue I saw.

PR #65, opened 2024-01-24
Add peak-metadata

willronchetti:
- Adds a new endpoint `/peek-metadata` that takes the same args as `/metadata` but returns the facet dictionary (currently just file size)

### DIFF
--- a/src/encoded/metadata.py
+++ b/src/encoded/metadata.py
@@ -158,7 +158,7 @@ def handle_metadata_arguments(context, request):
         return Response("Invalid parameters", status=400)
 
     if download_file_name is None:
-        download_file_name = 'metadata_' + datetime.utcnow().strftime('%Y-%m-%d-%Hh-%Mm') + '.tsv'
+        download_file_name = 'smaht_manifest_' + datetime.utcnow().strftime('%Y-%m-%d-%Hh-%Mm') + '.tsv'
 
     # Generate a header, resolve mapping
     # Note that this will become more complex as we add additional header types
--- a/src/encoded/metadata.py
+++ b/src/encoded/metadata.py
@@ -6,7 +6,7 @@ from snovault.search.search import (
     get_iterable_search_results, search
 )
 from snovault.search.search_utils import make_search_subreq
-from typing import Tuple
+from typing import Tuple, NamedTuple, List
 from urllib.parse import urlencode
 import csv
 import json
@@ -19,7 +19,7 @@ log = structlog.getLogger(__name__)
 
 
 def includeme(config):
-    config.add_route('peak_metadata', '/peak-metadata/')
+    config.add_route('peek_metadata', '/peek-metadata/')
     config.add_route('metadata', '/metadata/')
     config.add_route('metadata_redirect', '/metadata/{search_params}/{tsv}')
     config.scan(__name__)
@@ -28,6 +28,17 @@ def includeme(config):
 # For now, use enum code 0 for Files (this will expand greatly later to adapt support
 # for other types - Will 3 Jan 2023
 FILE = 0
+
+
+class MetadataArgs(NamedTuple):
+    """ NamedTuple that holds all the args passed to the /metadata and /peek-metadata endpoints """
+    accessions: List[str]
+    sort_param: str
+    type_param: str
+    include_extra_files: bool
+    download_file_name: str
+    header: Tuple[List[str], List[str], List[str]]
+    tsv_mapping: dict
 
 
 class TSVDescriptor:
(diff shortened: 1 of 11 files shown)


src/encoded/metadata.py
```diff
@@ -18,6 +19,7 @@
 
 
 def includeme(config):
+    config.add_route('peak_metadata', '/peak-metadata/')
```
drio18, line 22: I think peak should be peek here?

src/encoded/schemas/file.json
```diff
@@ -142,6 +142,24 @@
             "type": "string"
         }
     },
-    "facets": {},
+    "facets": {
+        "file_size": {
```
drio18, line 146: Can we exclude this facet? It wasn't desired for the benchmarking release.

src/encoded/metadata.py
```diff
@@ -159,6 +153,50 @@ def metadata_tsv(context, request):
     # Note that this will become more complex as we add additional header types
     header = generate_file_download_header(download_file_name)
     tsv_mapping = TSV_MAPPING[FILE]
+    return accessions, sort_param, type_param, include_extra_files, download_file_name, header, tsv_mapping
```
drio18, line 156: Consider creating a class/named tuple/dict to hold all these.

src/encoded/metadata.py
```diff
@@ -77,10 +81,10 @@ def read(self):
 
 def generate_file_download_header(download_file_name: str):
     """ Helper function that generates a suitable header for the File download """
-    header1 = ['###', 'Metadata TSV Download', '', '', '', '']
+    header1 = ['###', 'Metadata TSV Download', '', '', '', '', '']
     header2 = ['Suggested command to download: ', '', '',
                'cut -f 1 ./{} | tail -n +3 | grep -v ^# | xargs -n 1 curl -O -L '
```
drio18, line 86: Could we update this to use the annotated filename as the download name if it exists, otherwise falling back to the name from S3? My bash skills aren't good enough to suggest a solution, but would want to change the curl command to `curl -o {annotated_filename}` if it's not null.

src/encoded/tests/data/workbook-inserts/file_format.json
```diff
@@ -15,7 +15,8 @@
         "identifier": "BAM",
         "standard_file_extension": "bam",
 		"extra_file_formats": [
-			"BAI"
+			"BAI",
+            "VCF"
```
drio18, line 19: Consider reformatting

drio18 (commented):
Overall looks good. Left a couple of requests to consider updating the curl command to download with the annotated filename as well as to remove the size facet for now. If you're going to update the download here, will test via local deploy once that's in; otherwise can do so shortly.

src/encoded/metadata.py
willronchetti: haha good point!

Bianca-Morris (commented):
Hey, I tested this locally, and I ran into an issue where it keeps throwing 404s. I tried POSTing to `/peek_metadata/` and `/peek_metadata` both; neither worked. I created a new branch with this and the UI changes called `bm-wrr-metadata` if you'd like to take a look at it.
[Image from Gyazo](https://gyazo.com/c30e8bb0eb93cafac9bbe227159e2221)
(Note: it's currently set up to work with Doug's unmerged schema edits from another branch for columns and facets, so the only column you'll get is the selection one. If you select and then click download tsv that should still work, though.)

Bianca-Morris (approved):

src/encoded/metadata.py
```diff
@@ -159,38 +164,81 @@ def metadata_tsv(context, request):
...
+    if args.accessions:
+        search_param['accession'] = args.accessions
+    if args.sort_param:
+        search_param['sort'] = args.sort_param
+    search_param['limit'] = [1]  # we don't care about results, just the facets
+    search_param['additional_facet'] = ['file_size']
+    if args.include_extra_files:
+        search_param['additional_facet'].append('extra_files.file_size')
+    subreq = make_search_subreq(request, '{}?{}'.format('/search', urlencode(search_param, True)), inherit_user=True)
+    result = search(context, subreq)
```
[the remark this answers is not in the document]
willronchetti: Double check with Elizabeth but I do think we're opting to include them, though I think it is easy to change the default behavior as needed here if we decide differently either way

src/encoded/tests/data/workbook-inserts/file_format.json
Bianca-Morris: Seems like there's some debate as to whether a VCF is an "extra file format" on Slack? Not sure if it's relevant, but thought I'd let you know just in case it is: https://hms-dbmi.slack.com/archives/C051155U7QC/p1706202607799989

Bianca-Morris: Ah just saw your message! Ignore the above.

PR #69, opened 2024-01-29
Small bug fixes

willronchetti:
- Implement fallback for annotated field name, test it
- Repair extra files bug


drio18 (approved):
Looks good. Front-end is falling back to the `display_title` instead of `accession`, and I think same should probably be done here so the file name when downloaded is exactly as appears on the tables.

PR #71, opened 2024-01-31
Add user load command

willronchetti:
- Adds a command for loading (new) users into the portal from the OC list


src/encoded/commands/load_users_from_oc.py
```diff
@@ -0,0 +1,152 @@
...
+        """ Generates an email --> props mapping of users to post """
+        for u in user_csv_list:
+            user = self.build_user_from_row(u)
+            if user.email in self.user_dict:
+                raise UserCSVProcessorException(f'Found duplicate user in spreadsheet: {user.email}')
+            self.user_dict[user.email] = user
+        return self.user_dict
+
+    def remove_existing_users(self):
+        """ Strips out users who already have a user record """
```
dmichaels-harvard, line 84: I'd've just named this something like ignore_existing_users just because at first glance I'd thought this was going to remove the existing users from the database, rather than marking for ignore, the users in the list which are already in the database - though the comment obviously does say this.

dmichaels-harvard (approved):

PR #75, opened 2024-02-07
Add last_modified

willronchetti:
- Ensures `last_modified` is added to track edits (tested via local deploy)


drio18 (commented):
Do you think a test could be added to ensure all items have the `last_modified` field and is updated on PATCH? Worried someone, like me, will add an `_update()` method without the field on accident.

drio18 (approved):

PR #85, opened 2024-02-12
Cfm static section updates

crfmc:
Partly corresponds to [Trello Ticket 880](https://trello.com/c/2TSRUHWT/880-feedback-from-feb-1-release)
- Convert contents of file_uploads.jsx into an markdown file


Bianca-Morris (approved):
Looks good. One small note: I'd actually keep the JSX version of this file (`/docs/public/file_uploads.jsx`) since eventually we will have to embed a Youtube video into it (which will require switching back to JSX at some point in the future). No need to change anything in the inserts, though.

PR #92, opened 2024-02-14
Schema Updates for Data Submission

drio18:
This PR includes many schema changes to align with expected data coming our way for the benchmarking data submission. Breaking changes have been made and corresponding upgraders provided for properties currently populated on the dev or production portals.


aschroed (commented):
Not related to this PR but I notice that cell_ontology_id is required in cell_sample schema and the property uses a regular expression constraint - this could be an issue if the term is not in the CL ontology as yet (maybe irrelevant as presumably the only cell line data will be the benchmarking data and you already know the terms are available).

aschroed (approved):
Looks fine to me.

willronchetti (approved):
I have no issue with these changes, upgraders and tests look good to me

PR #93, opened 2024-02-15
Schema additions to facilitate automation

alexander-veit:
- Connect `MetaWorkflowRuns` with `FileSets`
- Add `role` to OutputFile

### DIFF
--- a/src/encoded/schemas/meta_workflow_run.json
+++ b/src/encoded/schemas/meta_workflow_run.json
@@ -79,6 +79,17 @@
         "cost": {
             "$merge": "encoded_core:schemas/meta_workflow_run.json#/properties/cost"
         },
+        "file_sets": {
+            "title": "File Sets",
+            "description": "File collections associated with this MetaWorkflowRun",
+            "type": "array",
+            "minItems": 1,
+            "uniqueItems": true,
+            "items": {
+                "type": "string",
+                "linkTo": "FileSet"
+            }
+        },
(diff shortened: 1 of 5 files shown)


src/encoded/schemas/output_file.json
```diff
@@ -93,6 +93,14 @@
         },
         "read_pair_number": {
             "$merge": "encoded:schemas/unaligned_reads.json#/properties/read_pair_number"
+        },
+        "role": {
```
drio18, line 97: Consider something more descriptive here. Maybe `output_status`?

drio18, line 97: Should be added to MWF `custom_pf_fields`: https://github.com/smaht-dac/smaht-portal/blob/main/src/encoded/schemas/meta_workflow.json#L120-L143.

src/encoded/schemas/meta_workflow_run.json
```diff
@@ -79,6 +79,17 @@
         "cost": {
             "$merge": "encoded_core:schemas/meta_workflow_run.json#/properties/cost"
         },
+        "file_sets": {
```
drio18, line 82: Consider adding a revlink on FileSet so the MWFRs appear there (?). Not sure if desirable.

drio18 (approved):
Looks good, dropping a few suggestions, but only the `custom_pf_fields` one is required.

PR #94, opened 2024-02-15
submits_for + restricted status 

willronchetti:
- Adds support for submits_for ie: restricting data submission/editing to particular users
- Adds support for restricted status, the same as released but looks for a special group `dbgap` (or `admin`) when executing a download


src/encoded/types/submission_center.py
```diff
@@ -17,3 +17,12 @@ class SubmissionCenter(Item):
     item_type = 'submission_center'
     schema = load_schema('encoded:schemas/submission_center.json')
     embedded_list = []
+
+    def __ac_local_roles__(self):
+        """This creates roles that the submission center item needs so it can be edited & viewed"""
+        roles = {}
+        sc_submitters = 'submits_for.%s' % self.uuid
+        roles[sc_submitters] = SUBMISSION_CENTER_SUBMITTER
+        sc_member = 'submission_centers.%s' % self.uuid
```
dmichaels-harvard, line 26: Just minor Python question: curious why using this % style string formatting rather than f-string?

dmichaels-harvard (approved):

src/encoded/types/acl.py
```diff
@@ -102,12 +109,14 @@
...
 ALLOW_OWNER_EDIT_ACL: Acl = [
-    (Allow, 'role.owner', ['edit', 'view', 'view_details']),
+    (Allow, OWNER_ROLE, ['edit', 'view', 'view_details']),
 ] + ONLY_ADMIN_VIEW_ACL
 
 # These two ACLs allow item editing
 SUBMISSION_CENTER_MEMBER_EDIT_ACL: Acl = [
-    (Allow, SUBMISSION_CENTER_RW, ['view', 'edit'])
+    # Note that only submitters can edit
+    (Allow, SUBMITTER_ROLE, ['edit']),
```
drio18, line 118: Should this include `view` as well? I believe a User can now upload things for a center to which they don't belong. May be useful to include some kind of validation on User to prevent this if it's an undesirable state.

src/encoded/types/file.py
```diff
@@ -163,9 +164,23 @@ def post_upload(context, request):
...
 
+def validate_user_has_protected_access(request):
+    """ Validates that the user who executed the request context either is
+        an admin or has the dbgap group
+    """
+    principals = request.effective_principals
+    if 'group.admin' in principals or 'group.dbgap' in principals:
+        return True
+    else:
+        return False
```
drio18, line 175: Consider simplifying.
```suggestion
    return False
```

src/encoded/types/file.py
```diff
@@ -163,9 +164,23 @@ def post_upload(context, request):
...
+
+
 @view_config(name='download', context=File, request_method='GET',
              permission='view', subpath_segments=[0, 1])
 def download(context, request):
+    if context.properties.get('status') == 'restricted' and not validate_user_has_protected_access(request):
+        raise HTTPForbidden('This is a restricted file not available for download without dbGAP approval, '
```
drio18, line 182:
```suggestion
        raise HTTPForbidden('This is a restricted file not available for download without dbGAP approval. '
```

src/encoded/authorization.py
```diff
@@ -113,8 +113,12 @@ def add_principal(principal):
 
     # SMaHT Specific stuff begins here (consortium and submission center)
     submission_centers = user_properties.get('submission_centers', [])
+    submits_for = user_properties.get('submits_for', [])
+    if submits_for:
+        principals.append('group.submitter')
+        principals.extend(f'submits_for.{submission_center}' for submission_center in submits_for)
     if submission_centers:
-        add_principal('role.submission_center_member_create')  # for add/create permissions
+        # add_principal('role.submission_center_member_create')  # for add/create permissions
```
drio18, line 121: Consider removing line.

src/encoded/tests/test_permissions.py
```diff
@@ -272,10 +290,11 @@ def test_submission_center_can_edit_file(test_submission_center, submission_cent
         'obsolete',
         'archived',
         'deleted',
-        'public'
+        'public',
+        'restricted'
     ])
     def test_submission_center_cannot_edit_file(test_submission_center, submission_center_user_app, released_file,
```
drio18, line 296: This test should probably be supplemented/duplicated for an app with a user of the same submission center but lacking the appropriate `submits_for` association to ensure no creation or edit.

src/encoded/types/file.py
```diff
@@ -163,9 +164,23 @@ def post_upload(context, request):
     return CorePostUpload(context, request)
 
 
+def validate_user_has_protected_access(request):
+    """ Validates that the user who executed the request context either is
+        an admin or has the dbgap group
```
drio18, line 169: Should this also allow people from the file's SubmissionCenter(s) to download? Not sure if that's desirable, though it seems intuitive; probably at least the person who uploaded the file should have perms to download it?

src/encoded/types/base.py
```diff
@@ -145,6 +145,9 @@ class Item(SnovaultItem):
         'released': acl.ALLOW_CONSORTIUM_MEMBER_VIEW_ACL,
         # Everyone can view - restricted to specific items via schemas.
         'public': acl.ALLOW_EVERYONE_VIEW_ACL,
+        # Intended to do additional computation to determine if download
+        # is allowed if it's a downloadable file, otherwise same as "released"
+        'restricted': acl.ALLOW_CONSORTIUM_MEMBER_VIEW_ACL,
```
drio18, line 150: This somewhat overlaps with the [existing property](https://github.com/smaht-dac/smaht-portal/blob/main/src/encoded/schemas/mixins.json#L278-L288) for access (that does nothing). Perhaps can turn it into a calculated property based on status now?

drio18 (commented):
Looks good. I think some of these changes warrant further testing, if I understand what's going on correctly. Also, looks like these changes obviate the necessity of a protected data consortium, as that permissioning has been relegated to the User groups now. Please correct me if that's wrong here.

src/encoded/types/base.py
willronchetti: Not sure I understand this comment?

drio18: For the benchmarking release, we added a dummy property on files `access_status` that indicates whether the file is designated open or restricted (I set it manually for the currently released files). Now that we have this status, may make sense to replace that property with a calculated property such that certain `status` values are correspond to`open` (such as `public` and `released`) and `restricted` corresponds to `Protected`. Maybe shouldn't implement this now and just keep as a dummy value until the dbgap stuff is in place; not sure on the timing.

willronchetti: Oh I see what you mean, I'll make this a calc-prop then as you suggest

src/encoded/schemas/user.json
```diff
@@ -65,6 +66,19 @@
         },
         "time_zone": {
             "$merge": "snovault:schemas/user.json#/properties/timezone"
+        },
+        "submits_for": {
+            "title": "Submits for",
+            "description": "Submission Centers a user is authorized to submit data for.",
+            "type": "array",
+            "permission": "restricted_fields",
+            "lookup": 70,
```
drio18, line 75: What does this do? I eliminated many of these in encoded-core and when porting from elsewhere; perhaps that was a mistake.

src/encoded/tests/test_permissions.py
```diff
@@ -607,6 +734,21 @@ def test_user_submission_center_consistency_alternate_id(testapp, test_submissio
...
+        consortia member can view it
+    """
+    atid = restricted_file['@id']
+    res = testapp.get(f'/{atid}@@download', status=307).json
+    assert 'smaht-unit-testing-wfout.s3.amazonaws.com' in res['message']
+    anontestapp.get(f'/{atid}@@download', status=403)
+    submission_center_user_app.get(f'/{atid}@@download', status=403)
+    testapp.patch_json(f'/{submission_center_user_app.extra_environ["REMOTE_USER"]}', {'groups': ['dbgap']})
+    res = submission_center_user_app.get(f'/{atid}@@download', status=307).json
+    assert 'smaht-unit-testing-wfout.s3.amazonaws.com' in res['message']
```
drio18, line 749: Consider removing as now redundant with `test_dbgap_group_with_restricted_status` above (or removing that test).

drio18 (approved):

src/encoded/schemas/user.json
willronchetti: `lookup` is used to order fields in the submission view (when you edit an item on the portal), but this is a carry over from fourfront and could probably be removed, especially if all the others have. We don't use the submission view a lot but it can be useful in certain situations. Not sure we need to add them back at this time.

src/encoded/types/file.py
aschroed: We do not allow that in 4DN - once a file is restricted status no one but admin can download it - however, use case may be different for SMaHT - we only set this status upon data release.

aschroed: is the dbgap group too general?  Is there additional control based on which study or does/will not that apply?

aschroed: Not for download but we do have a read-only-admin group but again not specific - anyone with that group can view - which is why I raised the question.
In 4DN if something is 'in review by lab' or 'uploaded' status then only a member of the Lab linked to a file/item can view/download it but that case is a bit different.

willronchetti: The assumption I've made here is that users who are submitting protected data would have the `dbgap` group, and that there is only one set of files for this purpose, so a single group suffices. This does definitely differ from 4DN, those with access to the protected data will be able to download it, while it is still viewable but not downloadable by those who do not have access.

aschroed: may be relevant and this is something we don't use a lot and is kind of a pain but we do have 'viewing_groups'  and Andrea needed a mechanism to share a non-released item with a select group of people who otherwise did not share a lab, award or other attribute so there is a mixin property called 'viewable_by' that is a string that is matched with a user.viewing_group and allows view of that item.  Could potentially be adapted for download but is clunky and there is likely a better way.  

Will I'm not sure I understand - is the eventual plan to have the restricted data elsewhere for download through dbgap so this group only applies to submitters?  Or if not will all the SMaHT data fall under the same dbgap study so as long as a user is approved for dbgap they can download any restricted file on the portal?

willronchetti: To my understanding, there will be 1 dbGaP study for SMaHT and it is binary choice whether they have it or not hence the single group. We will be storing this data ourselves and giving folks access to it through our portal, though I do think eventually moving data to SRA or ANViL was discussed previously as well.

PR #99, opened 2024-02-22
Submission Links

drio18:
This PR creates a new documentation page with links to collection searches useful for submitters when creating new submissions or desiring to view already submitted items.

Additionally, we include some basic facets, columns, rev-links, and calcprops to improve the searches above as well as for internal metadata navigation.


src/encoded/tests/data/master-inserts/user.json
```diff
@@ -298,7 +298,7 @@
     },
     {
         "email": "ozturkutku@gmail.com",
-        "uuid": "336b1174-8b79-4c39-a3ce-298be2cc1509",
+        "uuid": "1ade49a9-78e9-4d38-b127-92ac65598137",
```
willronchetti, line 301: Is this the uuid from prod?

willronchetti (approved):

src/encoded/tests/data/master-inserts/user.json
drio18: Prod and wolf; don't know why it was different here, but it was causing master inserts loading to crash.

willronchetti: Ok yes unifying with those envs is the right choice just wanted to be sure

PR #104, opened 2024-02-27
More Submission Alignment

drio18:
This PR further tweaks the data model for the transition to "automated" submission by users via:

- changing a property name on Analyte (upgrader included)
- removing a few unused properties
- adding admin permissions to select properties
- adding defaults and removing requirement for properties with only one possible enum

Additionally, we align inserts with the UUIDs currently existing on production environments to allow the deployment task action to load master inserts successfully.


src/encoded/upgrade/analyte.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+@upgrade_step("analyte", "1", "2")
+def upgrade_analyte_1_2(value: Dict[str, Any], system: Dict[str, Any]) -> None:
+    """Change `components` to `molecule_detail`."""
+    components = value.get("components")
+    if components is not None:
+        if components:
+            value["molecule_detail"] = components
+        del value["components"]
```
willronchetti, line 13: I get you're trying to be super safe here but just FYI
```
components = value.get("components")
if components:
    value["molecule_detail"] = components
    del value["components"]
```
is functionally equivalent since in the previous type definition you are ensuring there is at least 1 element in the list within the enum (since the empty list is the only non-null False value that would get you in a situation where you don't want to propagate the value but still need to del it).

src/encoded/tests/test_upgrade_analyte.py
```diff
@@ -0,0 +1,31 @@
...
+    "analyte,expected",
+    [
+        ({}, {"schema_version": "2"}),
+        (
+            {"components": [], "schema_version": "1"},
+            {"schema_version": "2"},
+        ),
+        (
+            {"components": ["Genomic DNA"], "schema_version": "1"},
+            {"molecule_detail": ["Genomic DNA"], "schema_version": "2"},
```
willronchetti, line 19: You might consider testing a length 2 array though is probably not necessary given how basic the upgrade is.

src/encoded/tests/test_upgrade_analyte.py
```diff
@@ -0,0 +1,31 @@
...
+from .utils import get_upgrader
+
+
+@pytest.mark.parametrize(
+    "analyte,expected",
+    [
+        ({}, {"schema_version": "2"}),
+        (
+            {"components": [], "schema_version": "1"},
+            {"schema_version": "2"},
```
willronchetti, line 15: This test is impossible no? The previous schema would throw a validation error if you had this.

willronchetti (approved):
Just needs version + CHANGELOG

src/encoded/tests/test_upgrade_analyte.py
drio18: Yup, goes back to the being safer than I need to be here; don't know why I did so. AFAIK, we don't have the means to test upgraders on items in the database properly (i.e. as it occurs on production), so we don't actually validate the test data. Would require validating against two different versions of the schema.

PR #114, opened 2024-03-13
ES Total Count, HealthView and RST Content Updates

utku-ozturk:
- Adds UI updates implemented in https://github.com/4dn-dcic/snovault/pull/284
- Ports HealthView page fixes previously [implemented](https://github.com/4dn-dcic/fourfront/commit/f2c2c6c42cb5cd3787b01062f607cbc9eef1960e) in 4DN for package-lock.json v3
- Adds admonition support for rst-to-html conversion in static content
- TODO: move `defaultExpanded` to the schema `page.json`

Screen Shot 2024-03-14 at 18 00 55


willronchetti (approved):
Just needs version bump

Bianca-Morris (approved):
Nice, looks good. We should think about how we can replicate this styling in other contexts (if you haven't done that already) since we might have JSX or HTML pages we want to use these things on in the future. Looks like some of this is already defined in a way we can use with HTML via the `.html-container`, but would be good to double check if this can be used/will work as expected for JSX, too, and how the markup needs to be structured, if so.

PR #118, opened 2024-03-14
Support common pipeline fields

alexander-veit:
This PR adds `common_fields` to MetaWorkflowRuns as well as `sequencing_center` to QualityMetric and WorkflowRun, so that it can be used as common field.


alexander-veit:
I tried to add 
```
"sequencing_center": {
      "$ref": "mixins.json#/sequencing_center/sequencing_center"
}
```
to file properties in order to avoid having to add the mixin to every file sub-type, but that did not resolve correctly.

drio18:
> I tried to add
> 
> ```
> "sequencing_center": {
>       "$ref": "mixins.json#/sequencing_center/sequencing_center"
> }
> ```
> 
> to file properties in order to avoid having to add the mixin to every file sub-type, but that did not resolve correctly.

You should be able to do this, but will require using the `$merge` referencing system instead of the `$ref` one.

drio18 (commented):
Looks good. I think your idea of moving the reference to File properties makes sense, but needs the other reference style, as commented.

My one request is to add a test to ensure all properties under MWFR `common_fields` are in fact present on the required items (WorkflowRun, OutputFile, and QualityMetric). Happy to lend a hand here if you'd like.

alexander-veit:
Good idea. I added a test.

alexander-veit:
I redeployed this to wolf. It works with `$merge`. Thanks for the tip.

PR #122, opened 2024-03-19
The extra_files branch with main and utk_es_max_hit branches merged in PLUS support for submission after server validation client timeout.

dmichaels-harvard:
* This is a temporary branch (extra_files_plus_main_20240314)
  which is Will's extra_files branch with main merged in (2024-03-14), and
  also Utku's utk_es_max_hit branch (PR-114) with documentation changes merge in.

* Added missing import of calculated_property from snovault to types/submitted_file.py.

* Documentation changes.

* Changes to support "resuming" smaht-sumbitr submission after a server
  validation "submission" timed out while waiting (via submit-metadata-bundle).
  **NOTE**: The only substantive changes for this are in **submission_folio.py**

  In this (server validation timeout) case the user can then run check-submission with
  the UUID for the validation submission, and if/when it is complete and successful,
  the user will be allowed to continue on to do the actual submission. Slightly tricky
  because the metadata file was uploaded (to S3) as a part of the validation submission,
  and/but when check-submission is run we don't want the user to have to specify this
  file again, partly because it is an odd user experience, but mostly because when we
  do the actual submission we want to make sure we use the EXACT file that was validated;
  and so to do this we grab the file from where it was uploaded as part of the validation
  submission (i.e. under an S3 key with the validation UUID) and copy it over to where
  it would normally be (i.e. under an S3 key with the submission UUID); and from there
  things continue as normal. Note also that both of the IngestionSubmission objects have
  a pointer to the other; i.e. the validation submission object has "submission_uuid"
  and the actual submission object has a "validation_uuid" (in the "parameters");
  this hookup is done by the smaht-submitr code.

  The "resuming" scare-quotes are because this is not really resuming a submission but
  rather resuming the process the submit-metadata-bundle was doing, i.e. where it does
  a server validation then then, if successful and okay with the user, it continues on
  to do the actual submission. The the "submission" scare-quotes for the server validation
  is because this is a submission in the sense that an IngestionSubmission object is
  created, but not an actual submission because it is a validate_only submission.

* Added display_title calculated property to IngestionSubmission to display either
  Validation or Submission depending on validate_only (followed by colon and uuid).


src/encoded/types/ingestion_submission.py
```diff
@@ -17,3 +20,23 @@ class IngestionSubmission(Item, SnovaultIngestionSubmission):
...
+    @calculated_property(schema=display_title_schema)
+    def display_title(
+        self,
+        request: Request,
+        parameters: Optional[dict] = None
+    ) -> str:
+        if isinstance(parameters, dict) and _to_bool(parameters.get("validate_only")) is True:
+            # TODO: Look into why validate_only is sometimes a bool and sometimes a string.
+            return f"Validation:{self.uuid}"
+        return f"Submission:{self.uuid}"
```
willronchetti, line 33: I would consider using something different from the UUID, you could use something like `Validation run for {filename} on {date}`?

src/encoded/types/ingestion_submission.py
```diff
@@ -17,3 +20,23 @@ class IngestionSubmission(Item, SnovaultIngestionSubmission):
...
     embedded_list = []
+
+    @calculated_property(schema=display_title_schema)
+    def display_title(
+        self,
+        request: Request,
+        parameters: Optional[dict] = None
+    ) -> str:
+        if isinstance(parameters, dict) and _to_bool(parameters.get("validate_only")) is True:
+            # TODO: Look into why validate_only is sometimes a bool and sometimes a string.
```
willronchetti, line 31: This is odd definitely, it should throw validation error no?

src/encoded/ingestion/submission_folio.py
```diff
@@ -95,3 +119,14 @@ def record_results(self, results: dict, validation_summary: list) -> None:
         # be very large, as it is an itemization of each/every object written to the database.
         #
         self.submission.process_standard_bundle_results(results, s3_only=True)
+
+    @staticmethod
+    def _construct_data_file_name_suitable_for_s3(filename: str) -> str:
```
willronchetti, line 124: Strongly recommend unit testing such a function, if not already tested in snovault

willronchetti (approved):
Some small comments on this, nothing major though

PR #123, opened 2024-03-21
feat: File Overview Page

crfmc:
File Overview Page giving additional details for File Items

- Upgrades SPC version to 0.1.80

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "@gmod/tabix": "^1.5.3",
         "@gmod/vcf": "^5.0.6",
         "@hms-dbmi-bgm/react-workflow-viz": "0.1.5",
-        "@hms-dbmi-bgm/shared-portal-components": "github:4dn-dcic/shared-portal-components#0.1.79",
+        "@hms-dbmi-bgm/shared-portal-components": "github:4dn-dcic/shared-portal-components#0.1.80",
         "auth0-lock": "^11.33.0",
         "d3": "^7.8.5",
         "date-fns": "^2.28.0",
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/item-pages/FileView.js
```diff
@@ -0,0 +1,173 @@
...
+                </ul>
+            </nav>
+            <h1 className="file-view-title-text">{context.display_title}</h1>
+        </div>
+    );
+};
+
+// Header component containing high-level information for the file item
+const FileViewHeader = (props) => {
+    const { context, session } = props;
```
Bianca-Morris, line 88: Might be good to do `context = {}` just so it has a default value and the line with selectedFile doesn't error out when indexing into context if it's not present. I don't think this should happen... but good to be safe.

src/encoded/static/components/item-pages/FileView.js
```diff
@@ -0,0 +1,173 @@
...
+                                </a>
+                                {i < arr.length - 1 ? (
+                                    <i className="icon icon-fw icon-angle-right fas"></i>
+                                ) : null}
+                            </li>
+                        );
+                    })}
+                </ul>
+            </nav>
+            <h1 className="file-view-title-text">{context.display_title}</h1>
```
Bianca-Morris, line 81: Might want to add optional chaining to this title, as you did the others.

Bianca-Morris (approved):
Tried to go through this one with a fine-toothed comb since it'll be deployed before the annual meeting. I made a bunch of notes, most of them quite small things, and mostly in the aim of trying to anticipate any potential crashes due to data loading, etc. This probably works fine in 99% of cases as-is, just being extra careful. It would be good if you could make these changes before merge, though.

src/encoded/static/components/item-pages/FileView.js
```diff
@@ -0,0 +1,173 @@
...
+                <div className="datum">
+                    <span className="datum-title">File Accession </span>
+                    <span className="vertical-divider">|</span>
+                    <span>
+                        <b className="accession">{accession}</b>
+                    </span>
+                </div>
+                <div className="datum right-group">
+                    <div className="status-group" data-status={status}>
+                        <i className="icon icon-circle fas"></i>
```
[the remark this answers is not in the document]
crfmc: I think aria-label is not necessary here, since it is non-interactive/static and not focusable. It does seem to lack some context without a proper "status" header, but it is read right after the file accession which I think does the job.

PR #125, opened 2024-03-26
Submitr ingestion status

dmichaels-harvard:
Changes to support tracking ingestion progess for smaht-submitr (via Redis).


development.ini.template
```diff
@@ -23,13 +23,15 @@ pyramid.default_locale_name = en
 # this line determines which load function is used in load_data
 # most deployments use: "load_test_data = snovault.loadxl:load_test_data"
 # note for smaht for now we only load_prod_data, extension can be done later
-load_test_data = snovault.loadxl:load_prod_data
+# load_test_data = snovault.loadxl:load_prod_data
+load_test_data = snovault.loadxl:load_local_data
```
willronchetti, line 27: Might want to revert this or add inserts you need

src/encoded/ingestion/loadxl_extensions.py
```diff
@@ -93,29 +100,93 @@ def package_loadxl_response(loadxl_response: Generator[bytes, None, None]) -> Di
...
+                progress_status[progress_status_enum] = 0
+        progress_status = {**progress_status,
+                           PROGRESS_LOADXL.TOTAL: total,
+                           PROGRESS_INGESTER.VALIDATION: validation}
+        def progress_tracker(progress: PROGRESS_LOADXL) -> None:  # noqa
+            nonlocal progress_status
+            def progress_message() -> None:  # noqa
+                # Just a convenience/courtesy so the consumer (smaht-submitr) doesn't have to cobble
+                # together a status message; but the data is still there of course if they want/need to.
+                nonlocal progress_status, total, validate_only
```
willronchetti, line 127: Is use of `nonlocal` necessary here? This entire block above as well is a little hard to follow

pyproject.toml
```diff
@@ -122,6 +122,7 @@ coverage = ">=6.2"
 codacy-coverage = ">=1.3.11"
 coveralls = ">=3.3.1"
 docutils = ">=0.16,<1"
+fakeredis = "^2.21.3"
```
willronchetti, line 125: I think `snovault` already provides a primitive for this? or perhaps it is in utils... either way I know there is another mechanism built into pytest that I think can work

willronchetti (approved):
Some small comments worth considering but nothing major

pyproject.toml
dmichaels-harvard: It complained without this.

PR #126, opened 2024-03-27
Dynamic Homepage

willronchetti:
- Make counts on the homepage dynamic with ability for easy future extension 
- Reorg tests such that workbook/non-workbook tests run in separate pytest invocations (interference sometimes occurring)


src/encoded/homepage.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def includeme(config):
+    config.add_route('home', '/home')
+    config.scan(__name__)
+
+
+def make_concurrent_search_requests(search_helpers):
+    """ Execute multiple search functions concurrently using a thread pool (since this is I/O bound). """
+    results = [-1] * len(search_helpers)  # watch out for -1 counts as indicative of an error
```
drio18, line 25: Are these -1 values being interpreted anywhere as indicative of an error currently?

src/encoded/homepage.py
```diff
@@ -0,0 +1,166 @@
...
+                print(f"Exception occurred in function {search_helpers[i][0].__name__}: {e}")
+    return results
+
+
+def extract_desired_facet_from_search(facets, desired_facet_name):
+    """ Grabs a single facet from a search response facets block """
+    for d in facets:
+        if d['field'] == desired_facet_name:
+            return d
+    return None
```
drio18, line 45: Consider logging inability to find facet name.

src/encoded/homepage.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def generate_colo829_cell_line_file_count(context, request):
```
drio18, line 48: Consider abstracting out the file count and assay count to separate functions to be reused for other datasets more easily.

drio18 (commented):
Nothing blocking, but strongly consider updating the searches to pull only those files front-end displays for the given cell lines to prevent required updates once things change. Can add the searches for all the other cell line collections pretty easily as well now so no need to update later to keep in sync with what's in the tables.

src/encoded/homepage.py
```diff
@@ -0,0 +1,166 @@
...
+
+def make_concurrent_search_requests(search_helpers):
+    """ Execute multiple search functions concurrently using a thread pool (since this is I/O bound). """
+    results = [-1] * len(search_helpers)  # watch out for -1 counts as indicative of an error
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        futures = []
+        for i, (func, kwargs) in enumerate(search_helpers):
+            future = executor.submit(func, **kwargs)
+            futures.append((i, future))
+        for i, future in futures:  # block in order so ordering is preserved
```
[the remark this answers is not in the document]
willronchetti: `future.result()` is a blocking call that will wait until a result is available to proceed - basically just ensuring ordering, though as you note there are smarter ways to do this but this mechanism right now is fairly generic

dmichaels-harvard (approved):

PR #127, opened 2024-03-27
File overview properties

drio18:
Here, we add new calculated properties to SubmittedFile and OutputFile types to capture information required for the overview page. Additionally, we add calcprops to separate out the various upstream items associated with files and create a new module (tentatively called `item_utils`) that centralizes common operations on items.


src/encoded/item_utils/cell_culture.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def is_cell_culture(properties: Dict[str, Any]) -> bool:
+    """Check if item is a cell culture."""
+    return item.get_type(properties) == "CellCulture"
+
+
+def get_cell_line(properties: Dict[str, Any]) -> Union[Dict[str, Any], str]:
+    """Get the cell line of a cell culture."""
+    return properties.get("cell_line", "")
```
willronchetti, line 13: Maybe add a similar `item.get_prop` as well?

src/encoded/item_utils/utils.py
```diff
@@ -0,0 +1,104 @@
...
+        self, identifier: str, collection: Optional[str] = None
+    ) -> Dict[str, Any]:
+        """Get item from request"""
+        return get_item_from_request(self.request, identifier, collection=collection)
+
+    def _get_item_from_auth_key(self, identifier: str) -> Dict[str, Any]:
+        """Get item from auth_key"""
+        return ff_utils.get_metadata(
+            identifier, key=self.auth_key, add_on="frame=object"
+        )
```
willronchetti, line 59: You are always able to use VirtualApp... I would not allow this personally as it is needlessly inefficient.

src/encoded/utils.py
```diff
@@ -18,3 +20,20 @@ def get_remote_user(item_with_environ: Union[Request, TestApp]) -> str:
 def get_environ_remote_user(environ: Dict[str, Any]) -> str:
     """Get remote user from associated environment."""
     return environ.get("REMOTE_USER", "")
+
+
+def get_item(
```
willronchetti, line 25: Reason for not just using this everywhere? We really should never use ff_utils in the portal

willronchetti (approved):
One comment I think merits addressing is removing use of ff_utils here, as you should always be able to use `VirtualApp` (or `request.embed`)

PR #129, opened 2024-04-02
Categories for RNASeq pipelines

drio18:
This PR updates category information to include RNASeq-related enums for:

- MetaWorkflow
- Software
- File (OutputFile + ReferenceFile)

### DIFF
--- a/src/encoded/schemas/file.json
+++ b/src/encoded/schemas/file.json
@@ -111,6 +111,7 @@
                     "Genome Region",
                     "Quality Control",
                     "Reference Genome",
+                    "RNA Expression",
                     "Sequencing Reads",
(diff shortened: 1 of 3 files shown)


src/encoded/schemas/file.json
```diff
@@ -111,6 +111,7 @@
                     "Genome Region",
                     "Quality Control",
                     "Reference Genome",
+                    "RNA Expression",
```
B3rse, line 114: I am not sure if that is a correct label, maybe "RNA Quantification" is better, or "Expression Quantification"? or we can think something else, see if Elizabeth as some other idea for this label

B3rse (changes requested):

PR #130, opened 2024-04-03
feat: format latest release date string

crfmc:
Formats latest release date above timeline data


Bianca-Morris (approved):
Looks good to me, although I haven't tested it locally. In the future, you might want to add some error fallback or retry code so that it doesn't appear to load forever if something goes wrong with the first load request.

PR #134, opened 2024-04-10
TPC metadata updates

drio18:
This PR makes significant changes to the metadata for donors and accompanying samples after discussions with the TPC over the past couple months as they've finalized information on their end.

Specifically, changes include:

- Addition of 2 new item types: FamilyHistory and MedicalTreatment
- Removal of 2 existing item types: Therapeutic (will not receive this info) and MolecularTest (moved to MedicalHistory since simpler data than expected)
- Removal of many properties from donor-related items the TPC will not be providing
- Addition of some properties to donor-related items the TPC will be providing
- Addition of enums for properties from the TPC
- Addition of `external_id` property to several item types to track identifiers from the TPC and other centers


src/encoded/schemas/death_circumstances.json
```diff
@@ -181,53 +134,52 @@
...
+            "enum": [
+                "Home",
+                "Hospital",
+                "Other"
+            ]
         },
-        "manner_of_death": {
-            "title": "Manner Of Death",
-            "description": "Manner of death of the donor",
+        "region_of_death": {
```
aschroed, line 147: This assumes US based donors, which may be what is expected but may want to add explicitly in description and if any possibility that not US then add an 'Other' option.

willronchetti (approved):
Made some points over slack but overall I don't have an issue with this in current form, looks like a natural extension to be expanded later. Needs version + CHANGELOG though.

src/encoded/schemas/mixins.json
```diff
@@ -239,6 +239,14 @@
             }
         }
     },
+    "external_id": {
```
aschroed, line 242: Curious as to why you did not use the already existing dbxrefs mixin?  There is a mechanism in fourfront via namespaces.json and some other magic that allows some dbxrefs to become links - that might be leveraged for some cases?

src/encoded/schemas/tissue_sample.json
```diff
@@ -69,22 +77,36 @@
         "schema_version": {
             "default": "1"
         },
-        "tissue_location": {
-            "title": "Tissue Location",
-            "description": "Original location of sample within source tissue",
+        "submitted_id": {
+            "pattern": "^[A-Z0-9]{3,}_TISSUE-SAMPLE_[A-Z0-9-_.]{4,}$"
+        },
+        "category": {
```
aschroed, line 83: Why did you add "category" mixin that is array type and then override - I think you need to remove the mixin?

aschroed (approved):
A few comments interspersed - one that I think is a mistake (tissue_sample category property) others not blocking.

I do have a concern about the way the medical_history item is set up with so many specific properties.  I get that this is how they want to provide the data from their system but worry that it may be subject to significant churn in the future and that may be a PITA for the front end team to use to generate a user friendly view if that might be needed.  Wonder if it might be a better idea to initially take this info in as a document and consider implementation details as things become clearer.

PR #135, opened 2024-04-11
Submission status page

alexander-veit:
This PR adds a Submission Status page for easier tracking of Filesets and associated MetaWorkflowRuns.

Demo: https://wolf.smaht.org/submission-status


src/encoded/schemas/mixins.json
```diff
@@ -199,6 +199,17 @@
             "uniqueKey": true
         }
     },
+    "comments": {
+        "comments": {
+            "title": "Comments",
+            "comment": "Intended for general comments concerning an item.",
```
willronchetti, line 205: You might want to indicate this is for internal use only, if that is indeed the intended use given permission = restricted_fields is set

src/encoded/submission_status.py
```diff
@@ -0,0 +1,148 @@
...
+        add_search_filters(search_params, filter)
+        subreq = make_search_subreq(
+            request, f"/search?{urlencode(search_params, True)}", inherit_user=True
+        )
+        search_res = search(context, subreq)["@graph"]
+        file_sets = []
+        for res in search_res:
+            file_set = res
+            file_set["submitted_files"] = process_files_metadata(res.get("files", []))
+            file_sets.append(file_set)
```
willronchetti, line 59: Is the reformatting really necessary? This is the only thing I see here that worries me, you have to get all requested results and post pass them here. It would probably be more efficient to use a search generator so you can stream results.

src/encoded/submission_status.py
```diff
@@ -0,0 +1,148 @@
...
+            "error": f'Error when trying to get submission status: {e}',
+        }
+
+    return {
+        "file_sets": file_sets,
+        "total_filesets": num_total_filesets,
+    }
+
+
+def add_search_filters(search_params, filter):
```
willronchetti, line 71: I would use more specific function names generally here since you are doing a very specific operation

src/encoded/submission_status.py
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+
+def search_total(context, request, search_params):
+    """Reads search params and executes a search total"""
+    ignored(context)
+
+    # Ensure this is always the case to prevent unintended slowness
+    if "limit" not in search_params:
+        search_params["limit"] = 0
```
willronchetti, line 142: This is probably not robust enough - one could still pass a bad "limit" parameter. You should ignore what is passed in for limit when getting a search total and always set it to 1.

src/encoded/static/components/static-pages/components/internal/SubmissionStatus.js
```diff
@@ -0,0 +1,604 @@
...
+            <React.Fragment>
+                <select
+                    className="custom-select"
+                    defaultValue="in review"
+                    onChange={(e) =>
+                        this.setFilter('fileset_status', e.target.value)
+                    }>
+                    <option value="all">All</option>
+                    <option value="in review">In Review</option>
+                    <option value="released">Released</option>
```
willronchetti, line 219: You probably want to include "public" and "restricted" - it may be easier to resolve the possible values from a facet in the backend and just allow all of them. You can get this from a standard (limit=0) search on files.

src/encoded/submission_status.py
```diff
@@ -0,0 +1,148 @@
...
+        "file_sets": file_sets,
+        "total_filesets": num_total_filesets,
+    }
+
+
+def add_search_filters(search_params, filter):
+    if not filter:
+        return
+    if "fileset_status" in filter and filter["fileset_status"] != "all":
+        search_params[STATUS] = filter["fileset_status"]
```
willronchetti, line 75: I like this sort of handling for simplicity but if you want to keep the UI the same for this you could map "released" in the UI to `["released", "restricted", "public"]` to ensure you pick up everything. Right now only `all` here will show what I think you would want to see under `released`.

src/encoded/schemas/mixins.json
alexander-veit: Yes, it is intended for internal use only. Modified the comment accordingly.

src/encoded/submission_status.py
alexander-veit: Good point

src/encoded/submission_status.py
alexander-veit: I see what you mean. Good idea. I will do that.

src/encoded/submission_status.py
alexander-veit: I am only getting 30 Filesets at a time (per POST request) and the items aren't very big. Loading of new filesets in the demo on wolf seems very quick. I agree that it could be more efficient, but it's probably not worth the effort? However, if you foresee issues with the current solution, I can optimize it of course.

willronchetti: Yeah let's punt on this for now

PR #140, opened 2024-04-15
Basecalling details

drio18:
This PR adds a new item type to capture the basecalling details required for ONT data.


src/encoded/schemas/basecalling.json
```diff
@@ -0,0 +1,104 @@
...
+            ]
+        },
+        "version": {
+            "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]+$"
+        },
+        "gpu": {
+            "title": "GPU",
+            "description": "GPU used for basecalling",
+            "type": "string"
+        },
```
willronchetti, line 79: You might consider making this suggested_enum at least so we don't need to reconcile differences in how these are entered

willronchetti (approved):
Needs version + CHANGELOG looks good otherwise though I think GPU (suggested)enum should be considered

PR #141, opened 2024-04-15
Link FileSet to Samples

drio18:
This PR adds an optional linkTo on FileSet to Samples for intended use for expected single cell datasets.

The issue this addresses is the following scenario:

- Many CellSamples (1/cell of the experiment) containing individual cell metadata
- No identifying information on the Analyte or Library items per cell
- Submitted files are demultiplexed per cell and need to be linked back to the appropriate CellSample

Since there is no data per single cell for Analyte and Library info, instead of forcing submission of 1 of these per cell, the above route instead requires individual FileSets per cell that can then be directly linked to the appropriate CellSample. An alternative method here would have been to allow this link on the File level rather than the FileSet, but this route was preferred to (1) reduce repetitive information if multiple files are provided per cell (e.g. BAM and variants) and (2) maintain FileSet as the unit of running pipelines should we need to do so in the future.


willronchetti (approved):
Good with this but rationale perhaps should be documented somewhere - also needs version and CHANGELOG

PR #144, opened 2024-04-15
Add parent samples to TissueSample

drio18:
This PR moves the `parent_samples` property on CellSample to mixins and adds it to TissueSample items to enable submitters to specify which samples derive from which.


willronchetti (approved):
No upgrader needed? Also needs version + CHANGELOG

PR #145, opened 2024-04-15
ReferenceGenome URL + build info

drio18:
This PR adds properties to ReferenceGenome to capture build info (RefSeq and GenBank IDs) as well as adding a URL prop to ReferenceFiles for pointers to the file download information (desired for reference files attached to genomes).


willronchetti (approved):
Needs version + CHANGELOG

PR #147, opened 2024-04-22
File Merge Group

willronchetti:
- Adds a calc prop to help determine if files in a file set are candidates for merge
- Adjusts access key automated expiration to 30 days


src/encoded/types/file_set.py
```diff
@@ -71,3 +81,93 @@ def meta_workflow_runs(self, request: Request) -> Union[List[str], None]:
             return result
         return
 
+    @staticmethod
+    def generate_sequencing_part(request, sequencing):
+        """ The sequencing part of the file_merge_group consists of the name of the sequencer,
+            the read type and the target read length - all required properties
```
alexander-veit, line 87: ...and flow cell

src/encoded/types/file_set.py
```diff
@@ -71,3 +81,93 @@ def meta_workflow_runs(self, request: Request) -> Union[List[str], None]:
...
+    @staticmethod
+    def generate_assay_part(request, library):
+        """ The library of the merge_file_group contains information on the assay
+            This basically just checks the assay code isn't a single cell type and if
+            it isn't return the identifier
+        """
+        assay = get_item_or_none(request, library.get('assay'))
+        assay_code = assay.get('code')
+        if assay_code in SINGLE_CELL_ASSAY_CODES:
+            return None
```
alexander-veit, line 104: Of course, the concern is that we generate incorrect group IDs, when we add a new single cell assay to the portal but forget to update `SINGLE_CELL_ASSAY_CODES` (which is likely to happen). Not sure how to best deal with that if we can't rely on search here though. A foursight check that fails when it detects a new single cell assay and reminds us that way?

alexander-veit: Or maybe it is better to maintain a list `NON_SINGLE_CELL_ASSAY_CODES ` and do 
```
if assay_code not in NON_SINGLE_CELL_ASSAY_CODES:
     return None
```
That way we won't generate wrong group Ids, but rather no group ID at all when we add new assay types and forget to update the list here.

alexander-veit (commented):
I think that looks good in general. I have, however, one comment concerning `SINGLE_CELL_ASSAY_CODES`.

src/encoded/types/file_set.py
willronchetti: Agreed!

src/encoded/types/file_set.py
```diff
@@ -72,3 +82,111 @@ def meta_workflow_runs(self, request: Request) -> Union[List[str], None]:
...
+
+    @staticmethod
+    def generate_assay_part(request, library):
+        """ The library of the merge_file_group contains information on the assay
+            This basically just checks the assay code isn't a single cell type and if
+            it isn't return the identifier
+        """
+        assay = get_item_or_none(request, library.get('assay'))
+        assay_code = assay.get('code')
+        if assay_code not in SINGLE_CELL_ASSAY_CODES:
```
alexander-veit, line 104: I don't think this works as expected.

docs/source/manifest.rst
```diff
@@ -0,0 +1,72 @@
...
+
+#. **File Accession** - this value is a server generated value specific to the SMaHT data portal, generally speaking it is unique with the exception that any associated extra files associated with the root file will have the same accession but different file extension. While you can reference files in data portal by this value, when communicating with others it is best to use the following field **File Name**.
+
+#. **File Name** - this value is the consortia-approved identifier for files, and is guaranteed to be a unique identifier encoding details about the file into the name. When downloading files, this will be the name of the file.
+
+#. **Size** - file size in bytes.
+
+#. **md5sum** - md5 of the file content.
+
+#. **Data Category** - this value tells you the file type, for example: Aligned Reads, Unaligned Reads or Variant Calls.
```
wfeng19, line 23: I think Variant Calls is actually a Data Type, so I would replace it with SNV or Indel.

docs/source/manifest.rst
```diff
@@ -0,0 +1,72 @@
...
+The preferred method to download data from the SMaHT data portal is to pass
+portal access credentials to the command provided in the ``smaht_manifest`` files.
+
+Most values in the manifest are single valued, but in some select scenarios multi-valued fields are separated by the pipe (``|``) character.
+
+Below are the columns listed in the ``manifest`` files as of the May 2024 data release.
+
+#. **File Download URL** - this URL, when navigated to while passing portal access credentials, will redirect to a pre-signed URL for downloading the associated file provided the user who made the request has access to the file.
+
+#. **File Accession** - this value is a server generated value specific to the SMaHT data portal, generally speaking it is unique with the exception that any associated extra files associated with the root file will have the same accession but different file extension. While you can reference files in data portal by this value, when communicating with others it is best to use the following field **File Name**.
```
wfeng19, line 15: Do users know what the difference between extra files and root files are? I might add an example saying that given a BAM file ``SMAFI557D2E7.bam``, the corresponding index file is ``SMAFI557D2E7.bam.bai``.

wfeng19 (commented):
I think it looks good, I just had a couple of comments on parts of the manifest documentation.

docs/source/manifest.rst
willronchetti: Although this is called `Data Category`, it is actually accessing the data type field, so would `Variant Calls` be correct? and do you think should be renamed to `Data Type`?

alexander-veit:
I think it works as expected.

PR #148, opened 2024-04-23
Submission Statistics

utku-ozturk:
Description coming soon...

Screen Shot 2024-05-16 at 16 54 01


src/encoded/visualization.py
```diff
@@ -19,15 +26,196 @@
...
+                    "interval": interval_to_es_interval[interval],
+                    "format": "yyyy-MM-dd"
+                }
+            }
+            if histogram_sub_aggs:
+                outer_date_histogram_agg[interval + '_interval_' + dh_field]['aggs'] = histogram_sub_aggs
+
+    search_param_lists['limit'] = search_param_lists['from'] = [0]
+    subreq          = make_search_subreq(request, '{}?{}'.format('/search/', urlencode(search_param_lists, True)) )
+    search_result   = perform_search_request(None, subreq, custom_aggregations=outer_date_histogram_agg)
```
willronchetti, line 173: You only need the aggregation result yes? Probably I should provide an endpoint to execute just an aggregation?

src/encoded/visualization.py
```diff
@@ -19,15 +26,196 @@
...
+    elif preset == 'last6months':
+        date_from = first_day_this_month - relativedelta(months=5)
+    elif preset == 'last12months':
+        date_from = first_day_this_month - relativedelta(months=11)
+    elif preset == 'thisyear':
+        date_from = datetime(today.year, 1, 1)
+    elif preset == 'previousyear':
+        date_from = datetime(today.year - 1, 1, 1)
+        date_to = datetime(today.year - 1, 12, 31)
+    elif preset == 'custom':
```
willronchetti, line 211: Stylistic comment - instead of doing it this way, you could write a function that builds a dictionary mapping your presets to the ``datetime`` values and just index into the dictionary in this function ie:

```
def generate_preset_table():
    today = datetime.today()
    first_day_this_month = today.replace(day=1)
    date_from, date_to = None, None
    d = {
        'thismonth': (first_day_this_month, None),
        'previousmonth': (first_day_this_month - relativedelta(months=1), first_day_this_month - relativedelta(days=1))
        ...
     }
     return d
```

You can make this even cleaner by using a `NamedTuple` for `date_from` and `date_to`, but a standard tuple will work as well.

willronchetti (approved):
Backend generally looks fine to me, but please create a ticket for aggregation endpoint so we can make it more efficient

PR #149, opened 2024-04-25
File sample summary

drio18:
This PR adds a new calcprop to files to capture the sample information to be shown on the file overview page.

To do so, we add new methods as needed in `item_utils` (+ refactor existing ones) and refactor some `utils` modules to share code from testing (+ enable virtual/test app use with the `item_utils` modules).


src/encoded/tests/test_item_utils_sample.py
```diff
@@ -0,0 +1,120 @@
...
+from ..item_utils import item
+from ..item_utils.sample import (
+    get_sample_descriptions,
+    get_sample_names,
+    get_studies,
+)
+from ..item_utils.utils import RequestHandler
+
+
+@pytest.mark.workbook
```
willronchetti, line 16: FYI if the entire file is workbook tests you can put 

```pytestmark = [pytest.mark.workbook]```

at the top of the file and it will automatically mark all tests in the file as workbook tests.

File: src/encoded/utils.py [hunk 2]
```diff
@@ -92,3 +99,102 @@ def get_configuration_value(property_name: str,
...
+    resource_path = get_formatted_resource_path(
+        identifier, collection=collection, add_on=add_on
+    )
+    response = testapp.get(resource_path, status=status)
+    if response.status_int == 301:
+        return response.follow().json
+    return response.json
+
+
+def get_frame_add_on(frame: Union[str, None]) -> str:
```
willronchetti, line 122: You might want to validate that `frame` is one of `raw`, `object` or `embedded` as to catch errors early

src/encoded/utils.py
```diff
@@ -92,3 +99,102 @@ def get_configuration_value(property_name: str,
...
+
+def get_kebab_formatted_collection(collection: str) -> str:
+    """Format collection to  kebab case.
+
+    Intended for collection to come in camel- or snake-case.
+    """
+    return to_snake_case(collection).replace("_", "-")
+
+
+def pluralize_collection(collection: str) -> str:
```
willronchetti, line 175: Does this method not work? https://github.com/4dn-dcic/utils/blob/master/dcicutils/lang_utils.py#L114

willronchetti (approved):
Needs version + CHANGELOG

PR #154, opened 2024-05-02
Test change swapping WF's first and last name.

wfeng19:
Test change swapping WF's first and last name.


src/encoded/tests/data/master-inserts/user.json
```diff
@@ -344,8 +344,8 @@
             "9626d82e-8110-4213-ac75-0a50adf890ff"
         ],
         "time_zone": "US/Eastern",
-        "last_name": "Feng",
-        "first_name": "William"
+        "last_name": "William",
+        "first_name": "Feng"
```
drio18, line 348: I don't know about this...

PR #156, opened 2024-05-03
Extend release script (obsolete files)

alexander-veit:
Allow release script to pass in an identifier for an obsolete file. Can be used if the new file to release replaces a previously released file.


src/encoded/commands/release_file.py
```diff
@@ -100,7 +101,9 @@ def __init__(self, auth_key: dict):
         self.warnings = []
         self.file_accession = ""
 
-    def prepare(self, file_identifier: str, dataset: str) -> None:
+    def prepare(
+        self, file_identifier: str, obsolete_file_identifier: str, dataset: str
```
drio18, line 105: Consider passing the new arg as a kwarg since not required.

drio18 (approved):

PR #157, opened 2024-05-03
Add paired_with to output files

alexander-veit:
I tried to add the rule
```
"allOf": [
        {
            "properties": {
                "read_pair_number": {
                    "not": {
                        "const": "R2"
                    }
                }
            },
            "not": {
                "required": [
                    "paired_with"
                ]
            }
        }
    ],
```
but that did not work. It does not allow anthing marked as R2. I removed the validation entirely.


drio18 (approved):
Hmm, not sure why the validation isn't working, but all looks fine and we're the only ones that can add the field for now. Still think it's worth getting that validation to work, and happy to take a stab at it myself, but don't think it should block this.

PR #161, opened 2024-05-07
feat: file download doc page

crfmc:
Corresponds to [Trello Ticket 267](https://trello.com/c/oXvFbbs9/267-final-changes-to-the-how-to-download-files-doc)

Deployed on [smaht-wolf](https://wolf.smaht.org/docs/how-to-download)


src/encoded/tests/data/master-inserts/page.json
```diff
@@ -373,6 +374,24 @@
...
+            "enabled": true
+        },
+        "status": "public",
+        "consortia": [
+            "358aed10-9b9d-4e26-ab84-4bd162da182b"
+        ],
+        "submission_centers": [
+            "9626d82e-8110-4213-ac75-0a50adf890ff"
+        ]
+    },
```
Bianca-Morris, line 394: Just a heads up, there was already a page available for this `859b2c48-0bda-406c-b49b-07d68bd236d4`. If we're going with this one, please delete the other one from this file.

docs/source/file_download.rst
```diff
@@ -0,0 +1,76 @@
+=================================
+File Download on the Data Portal
```
Bianca-Morris, line 2: One thing with this: look closely at the document once this is rendered. Sometimes our RST renderer eats spaces. I think the way you've done this here is probably fine, since you're keeping to a single line continuous text/paragraphs. But it's always good to double check it, since it doesn't appear the same way on local as it does on live/prod envs.

Bianca-Morris (approved):
Approve with one actionable note.

PR #162, opened 2024-05-08
Broaden `derived_from` to all files

drio18:
This PR allows any file type as a link on the `derived_from` prop, specifically to enable links to OutputFiles (as desired for the in silico BAMs).

In addition, to allow this to work properly on submission, we change the File `unique_key` to `submitted_id` from `accession`. This is required to ensure submitted IDs are properly found on look up with the broader scope above.

### DIFF
--- a/src/encoded/types/file.py
+++ b/src/encoded/types/file.py
@@ -347,7 +347,7 @@ def _build_file_embedded_list() -> List[str]:
 
 @abstract_collection(
     name="files",
-    unique_key='accession',
+    unique_key="submitted_id",  # To permit lookup on submission
     properties={
         "title": "Files",
         "description": "Listing of Files",
--- a/src/encoded/item_utils/item.py
+++ b/src/encoded/item_utils/item.py
@@ -69,3 +69,8 @@ def get_external_id(properties: Dict[str, Any]) -> str:
 def get_tags(properties: Dict[str, Any]) -> List[str]:
     """Get tags from properties."""
     return properties.get("tags", [])
+
+
+def get_submitted_id(properties: Dict[str, Any]) -> str:
+    """Get submitted ID from properties."""
+    return properties.get("submitted_id", "")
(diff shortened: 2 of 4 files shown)


src/encoded/types/base.py
```diff
@@ -66,18 +67,20 @@ def __init__(self, *args, **kw):
...
         """
         resource = super(AbstractCollection, self).get(name, None)
         if resource is not None:
             return resource
+        if is_accession(name):
+            resource = self.connection.get_by_unique_key('accession', name)
+            if resource is not None:
+                if not self._allow_contained(resource):
+                    return default
+                return resource
```
willronchetti, line 83: Did you find this to be necessary? I think L75 should handle this as it calls out to https://github.com/4dn-dcic/snovault/blob/master/snovault/resources.py#L309-L317 no? If this doesn't work, I think it means the unique key is not specified correctly?

src/encoded/types/file.py
```diff
@@ -347,7 +347,7 @@ def _build_file_embedded_list() -> List[str]:
 
 @abstract_collection(
     name="files",
-    unique_key='accession',
+    unique_key="submitted_id",  # To permit lookup on submission
```
willronchetti, line 350: Is this change why the above fix is needed for accession?

willronchetti (approved):
Small questions nothing major

src/encoded/types/base.py
drio18: This is helpful to get over the limitation of the designation of only one unique key, limiting lookups with the collection (e.g. `/file/.../`) to either the UUID or the unique key. For example, for all submitted files, the unique key is the `submitted_id`, and that means accessions currently do not work when provided with the collection. The above fixes this to allow accessions, which should always work given the accession alone is able to redirect to the item.

More importantly, during submission with submitr links are formatted into the collection-style lookup (`/file/.../`), so accessions are not able to be used currently for submitted items and other item types that have accessions but non-accession unique keys. With the change above, accessions are usable, as well as UUIDs and designated unique keys.

src/encoded/types/file.py
drio18: This was necessary to allow the use of `submitted_id` for generic File linkTos. It exposed the accession problem to me, but isn't the sole reason for the fix.

PR #165, opened 2024-05-08
Update preparation items

drio18:
This PR updates the Preparation items to be made available for manual submission.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,16 @@ smaht-portal
 Change Log
 ----------
 
+0.52.0
+======
+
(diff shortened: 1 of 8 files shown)


wfeng19 (commented):
I think everything looks good. I did notice that we have the "description" property on SamplePrep but not on any of the other preparation items. For consistency purposes, I'm not sure if we want to remove "description" from SamplePreparation or add it to LibraryPrep and AnalytePrep.

drio18:
> I think everything looks good. I did notice that we have the "description" property on SamplePrep but not on any of the other preparation items. For consistency purposes, I'm not sure if we want to remove "description" from SamplePreparation or add it to LibraryPrep and AnalytePrep.

Good catch, will add this in to all of them.

PR #166, opened 2024-05-10
Download Cli

willronchetti:
- Adds a new API, `download_cli`, that accepts any unique key/resource path and returns AWS Access keys for download. It does this by first checking access against `@@download`, then processing the redirect result if successful and calling get_federation_token subsequently.
- TODO: hook in new version of encoded-core, which will change response to include "download" instead of "upload" credentials, and constrain the permissions to only download

### DIFF
--- a/src/encoded/download_cli.py
+++ b/src/encoded/download_cli.py
@@ -1,0 +1,64 @@
+from json import JSONDecodeError
+from pyramid.view import view_config
+from pyramid.response import Response
+from pyramid.httpexceptions import HTTPTemporaryRedirect
+from urllib.parse import urlparse
+from snovault.util import debug_log
+from dcicutils.misc_utils import ignored
+from encoded_core.types.file import external_creds
+from structlog import getLogger
+
+
+log = getLogger(__name__)
+
+
+def includeme(config):
+    config.add_route('download_cli', '/download_cli/')
+    config.scan(__name__)
+
+
+def extract_bucket_and_key(url):
+    """ Takes an HTTPS URL to S3 and extracts the bucket and key """
+    parsed_url = urlparse(url)
+    bucket = parsed_url.netloc.split('.')[0]
+    key = parsed_url.path.strip('/')
+    if '?' in key:  # remove query params
+        key = key.split('?', 1)[0]
+    return bucket, key
+
+
+def generate_creds(e):
+    """ Processes HTTPTemporary redirect response from the app, grabbing the bucket/key
+        and calling external_creds with upload=False
+    """
+    url = e.location
+    bucket, key = extract_bucket_and_key(url)
+    return external_creds(bucket, key, name=f'DownloadCredentials')
+
+
+@view_config(route_name='download_cli', request_method=['GET', 'POST'])
+@debug_log
+def get_download_federation_token(context, request):
+    """ Runs the external_creds function with upload=False assuming user passes auth check """
+    ignored(context)
+    # TODO: refactor to helper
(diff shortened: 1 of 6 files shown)


src/encoded/tests/test_download_cli.py
```diff
@@ -0,0 +1,97 @@
...
+        'item': f'{uri}'
+    }).json['upload_credentials']
+    assert 'AccessKeyId' in res
+    assert 'ASIA' in res['AccessKeyId']  # only real check we can do that this key is real
+    assert 'SecretAccessKey' in res
+    assert 'SessionToken' in res
+
+
+@pytest.mark.workbook
+def test_download_cli_workbook_post(workbook, es_testapp):
```
drio18, line 43: Consider adding a test case for behavior when provided an item that lacks an `@@download` endpoint.

src/encoded/download_cli.py
```diff
@@ -0,0 +1,66 @@
...
+        atid = request.GET.get('item')
+        if not atid:
+            return Response('No item query parameter present', status=400)
+
+    # Direct to @@download to track via GA and run perm check
+    if '@@download' in atid:  # allow direct pass
+        try:  # this call will raise HTTPTemporary redirect if successful
+            request.embed(f'{atid}', as_user=True)
+        except HTTPTemporaryRedirect as e:
+            return generate_creds(e)
```
drio18, line 60: This method of getting the creds is a little indirect and is dependent on the workings of `@@download.` Consider more direct alternatives to getting this data, such as by passing args to `@@download`.

drio18 (approved):
Nothing blocking, but would recommend moving this to encoded-core rather than this repo to enable sharing with other portals.

src/encoded/tests/test_download_cli.py
```diff
@@ -0,0 +1,97 @@
...
+    _test_uri_post(es_testapp, f'{atid}@@download')
+    _test_uri_post(es_testapp, f'/{uuid}/@@download')
+    _test_uri_post(es_testapp, f'/{accession}/@@download')
+    # test without @@download
+    _test_uri_post(es_testapp, f'{atid}')
+    _test_uri_post(es_testapp, f'/{uuid}')
+    _test_uri_post(es_testapp, f'/{accession}')
+    # test extra file 1 with @@download
+    # NOTE: dummy data that doesn't use the accession as the file name
+    # does NOT work to download!
```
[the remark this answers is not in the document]
willronchetti: Correct yes

docs/source/download_cli.rst
```diff
@@ -0,0 +1,23 @@
+======================================================
+Downloading Data from SMaHT Manifest files with awscli
+======================================================
+
+
+In addition to direct processing of file downloads via ``curl`` in the manifest file, DAC also supports downloads through the ``awscli``. In some situations, you may get better bandwidth using the ``awscli`` over ``curl`` or equivalent. We recommend testing both to see which ends up better.
+
+The ``download_cli`` API will return short term AWS credentials for direct access to DAC buckets to facilitate data transfers over ``awscli``. When downloading the manifest file, you may specify you want download URLs in this format. The standard command provided with the manifest will not work though, you must use a different command. Note that this command assumes you have previously installed ``awscli``. It will work with both version 1 and 2 of ``awscli``. It also uses ``jq`` to process the JSON response.
```
dmichaels-harvard, line 8: maybe comment: jq can be easily installed on MacoS via brew install jq

dmichaels-harvard (approved):

PR #171, opened 2024-05-15
Bm docs nav update

Bianca-Morris:
**Changelog:**
- Add a bunch of new inserts to serve as the new links to these various pages
- Rework the old inserts into redirect-only pages, so that people who have bookmarked old links don't lose their place
- Permission and order tweaks for sanity's sake

**Notes:**
~There seems to be a bug with redirects at the moment. Not sure what would have affected this; I tried it on staging before I did this and it worked fine. More discussion on this will be on the ticket. (Also, I could be doing something wrong... not sure.)~ Looks like it was an issue with redirect location being cached by the browser. A hard refresh was not enough to resolve it; needed to disable cache, and re-load with `target: "/<path>"` in the redirect object.

**Related ticket:**
https://trello.com/c/GjtQDsFc/268-update-documentation-navbar-tab-add-redirects

**Screenshot:**
[Image from Gyazo](https://gyazo.com/0c4324d594bc9d015fd3ac067f47af27)


Bianca-Morris:
[Update Documentation Navbar Tab & Add Redirects](https://trello.com/c/GjtQDsFc/268-update-documentation-navbar-tab-add-redirects)

crfmc (approved):

PR #172, opened 2024-05-16
Submission Status page: More filters & add file group information

alexander-veit:
- Add filters for CellLine and CellCultureMixture/Tissue
- Some refactoring of the React component
- Color filesets by file group


src/encoded/submission_status.py
```diff
@@ -57,10 +58,31 @@ def get_submission_status(context, request):
...
+
+            if "file_group" in file_set:
+                fg = file_set["file_group"]
+                fg_str = f"{fg['submission_center']}_{fg['sample_source']}_{fg['sequencing']}_{fg['assay']}"
+                file_set["file_group"] = fg_str
+                # Place holder that will be replaced in the next step
+                file_group_color_map[fg_str] = None
+            else:
+                file_set["file_group"] = "No file group assigned"
+                file_set["file_group_color"] = "#eeeeee"
```
willronchetti, line 74: I'd strongly prefer you create a top level "color map" that maps colors to their IDs here so it is more clear what colors you're using where.

src/encoded/submission_status.py
```diff
@@ -122,6 +146,20 @@ def add_submission_status_search_filters(
         search_params["libraries.assay.display_title"] = filter["assay"]
     if "sequencer" in filter and filter["sequencer"] != "all":
         search_params["sequencing.sequencer.display_title"] = filter["sequencer"]
+    if "cell_line" in filter and filter["cell_line"] != "all":
+        search_params["libraries.analytes.samples.sample_sources.cell_line.code"] = (
+            filter["cell_line"]
+        )
+        search_params["libraries.analytes.samples.sample_sources.code"] = (
+            "No+value"  # Exlude mixtures from results
+        )
```
willronchetti, line 155: I think subsequent call to `urlencode` makes use of `+` unnecessary here, does this work? Also typo: "exclude"

willronchetti (approved):

PR #173, opened 2024-05-21
Multiple analytes per library

drio18:
This PR introduces a breaking change to the data model, changing the existing Library link to an Analyte to an array of links and pluralizing the property name accordingly.

Accompanying changes include:

- Library upgrader for existing items
- `item_utils` updates
- Embed updates
- File calcprop refactoring
- FileSet calcprop refactoring (switched over to using `item_utils`)
- Command refactoring for file naming and file release
- Test + workbook inserts updates
- Submission status updates (courtesy of Alex)

PR #176, opened 2024-05-22
Bm benchmarking v3

Bianca-Morris:
**Changelog:**
- Add benchmarking tables for primary tissue data, organized by Donor, then Tissue type
   - Add new inserts/remove old ones
- Update navbar and navbar links

**Review Notes:**
- @drio18 It would be helpful if you could double check the table URLs are correct

**Deployment Notes:**
- As of May 22nd at 6:17pm, this is deployed on smaht-wolf

### DIFF
--- a/src/encoded/static/components/static-pages/components/benchmarking/BenchmarkingDataMap.js
+++ b/src/encoded/static/components/static-pages/components/benchmarking/BenchmarkingDataMap.js
@@ -183,7 +183,7 @@ export const BenchmarkingDataMap = {
     },
     Donor1: {
         navBarTitle: 'Donor ST001',
-        title: 'Donor ST001 : Primary Tissue Data',
+        title: 'Donor ST001 : Benchmarking Tissue Data',
         description: primaryTissuePageDescription,
         type: 'Primary Tissue Data',
         path: '/data/benchmarking/donor-st001',
@@ -204,7 +204,7 @@ export const BenchmarkingDataMap = {
     },
     Donor2: {
         navBarTitle: 'Donor ST002',
-        title: 'Donor ST002 : Primary Tissue Data',
+        title: 'Donor ST002 : Benchmarking Tissue Data',
         description: primaryTissuePageDescription,
         type: 'Primary Tissue Data',
         path: '/data/benchmarking/donor-st002',
@@ -225,7 +225,7 @@ export const BenchmarkingDataMap = {
     },
     Donor3: {
         navBarTitle: 'Donor ST003',
-        title: 'Donor ST003 : Primary Tissue Data',
+        title: 'Donor ST003 : Benchmarking Tissue Data',
         description: primaryTissuePageDescription,
         type: 'Primary Tissue Data',
         path: '/data/benchmarking/donor-st003',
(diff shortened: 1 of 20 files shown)


drio18:
The search URLs will have to be updated once [this PR](https://github.com/smaht-dac/smaht-portal/pull/173) goes in, which should happen shortly, so I'll provide the new URLs once merged.

crfmc (approved):

src/encoded/static/components/static-pages/components/benchmarking/BenchmarkingDataMap.js
```diff
@@ -181,130 +181,74 @@ export const BenchmarkingDataMap = {
...
+        path: '/data/benchmarking/donor-st001',
         tabMapArray: [
             {
-                eventKey: '#donor-1',
-                title: 'Donor 1',
-                // searchHref: '/search/?type=File&status=released&status=restricted&status=public',
+                eventKey: '#liver',
+                title: 'Liver',
+                searchHref:
+                    '/search/?type=File&status=released&status=restricted&status=public&file_sets.libraries.analyte.samples.sample_sources.code=ST001-1A',
```
drio18, line 195: I've merged my branch, so the update below is required for all these searches:

```suggestion
                    '/search/?type=File&status=released&status=restricted&status=public&file_sets.libraries.analytes.samples.sample_sources.code=ST001-1A',
```

src/encoded/static/components/static-pages/components/benchmarking/BenchmarkingDataMap.js
```diff
@@ -181,130 +181,74 @@ export const BenchmarkingDataMap = {
...
         type: 'Primary Tissue Data',
-        path: '/data/benchmarking/colon',
+        path: '/data/benchmarking/donor-st003',
         tabMapArray: [
             {
-                eventKey: '#donor-1',
-                title: 'Donor 1',
-                // searchHref: '/search/?type=File&status=released&status=restricted&status=public',
+                eventKey: '#brain',
+                title: 'Brain',
```
drio18, line 235: Any reason why we're not providing the searches for the brain tissues here and below? Relevant codes are `ST003-1Q` and `ST004-1Q`, resp.

drio18 (commented):
See comments for required updates for the recent data model change.

src/encoded/static/components/static-pages/components/benchmarking/BenchmarkingDataMap.js
Bianca-Morris: Since we won't have those right away, Elizabeth wants to display "coming soon". And then we'll update later with the real URLs.

src/encoded/static/components/static-pages/components/benchmarking/BenchmarkingDataMap.js
Bianca-Morris: Thanks for this reminder. Went ahead and made that change.

Bianca-Morris:
Heads up to anyone looking at this on or after May 28th, we are commenting out the URLs because the primary tissue data is not yet ready. This will have the UI render "coming soon" instead.

PR #181, opened 2024-06-09
Statistics Follow-up

utku-ozturk:
### Filtering
- Previous 60 days added to date range options
- Cumulative sum checkbox added to switch between individual bars and cumulative sum

Screen Shot 2024-06-12 at 17 46 24

### File Downloads
- File downloads splitted into two charts (count and volume)
- Assay Type and Dataset views added

Screen Shot 2024-06-12 at 17 43 48

### File Views
- Assay Type and Dataset views added

Screen Shot 2024-06-12 at 17 49 04

### Page Views
-  Page views, Uniqe users by country/city views added
- Page title and URL views added

Screen Shot 2024-06-12 at 17 50 32

### Misc
- tracking_item schema fields that are still mapped to Universal Analytics naming convention are renamed/replaced
- glitches in mobile view fixed

### Next?
- We'll purge the existing tracking items in production. Then we'll run the GA4-to-TrackingItem conversions to recreate new tracking items after this PR is merged and deployed.


utku-ozturk:
[/statistics submission & usage tabs follow-ups](https://trello.com/c/WJdqZW3e/282-statistics-submission-usage-tabs-follow-ups)

src/encoded/static/components/viz/AreaChart.js
```diff
@@ -566,10 +566,16 @@ export class HorizontalD3ScaleLegend extends React.Component {
         return false;
     }
 
+    componentDidUpdate(pastProps, pastState) {
+        setTimeout(function () {
+            ReactTooltip.rebuild();
+        }, 1000);
+    }
```
Bianca-Morris, line 573: Is there any way to configure this delay so that it's not hard-coded? Is there a particular part of the interface that is changing in response to a change in props or state that is resulting in the necessity to rebuild the tooltips? If so, using a conditional here instead of a timeout will probably be better in the long run.

src/encoded/static/components/viz/AreaChart.js
```diff
@@ -710,7 +716,8 @@ export class AreaChart extends React.PureComponent {
     }
 
     static isTermUrl(term){
-        return term && typeof term === 'string' && term.length > 3 && term.charAt(0) === '/' && term[term.length - 1] === '/';
+        // return term && typeof term === 'string' && term.length > 3 && term.charAt(0) === '/' && term[term.length - 1] === '/';
```
Bianca-Morris, line 719: Is keeping this here helpful? Is there a chance it will be used in the future?

src/encoded/static/components/static-pages/components/StatisticsPageViewBody.js
```diff
@@ -228,63 +228,94 @@ export const commonParsingFxn = {
...
             if (typeof topCount === 'number' && topCount > 0) {
-                currItem.children = _.sortBy(currItem.children, (item) => -1 * item.total).slice(0, topCount);
+                currentItem.children = _.sortBy(currentItem.children, (item) => -1 * item.total).slice(0, topCount);
             }
 
-            return currItem;
+            return currentItem;
 
-        }).reverse(); // We get these in decrementing order from back-end
+        }); // We get these in decrementing order from back-end
```
Bianca-Morris, line 316: Is this comment still true?

Bianca-Morris (approved):
I have a few small comments, but nothing preventing merge. I will say, `src/encoded/static/components/static-pages/components/StatisticsPageViewBody.js` is getting pretty long, and it might be good to try to figure out a way to separate that out into multiple files in the near future. Nice job on this; I know it's been a while in the making.

src/encoded/static/components/static-pages/components/StatisticsPageViewBody.js
utku-ozturk: yeah, definitely should not be there anymore!

src/encoded/static/components/viz/AreaChart.js
utku-ozturk: good catch! let me check it. thanks!

willronchetti (approved):

src/encoded/static/components/viz/AreaChart.js
utku-ozturk: @Bianca-Morris  I tested it once more. The `shouldComponentUpdate(nextProps, nextState)` life-cycle function prevents unnecessary `componentDidUpdate` calls, so the `ReactTooltip.rebuild` is called when only the color scale or terms change.

PR #184, opened 2024-06-13
SN Update submission schema from TPC and GCC feedback

sarahgonicholson:
- Remove SamplePreparation item and merges the property homogenization_method with AnalytePreparation
- Add properties to Analyte based on GCC Benchmarking Extraction input [here](https://docs.google.com/spreadsheets/d/1_agfKxRFgBZtUOrYLgIwBk1pmUqzwKxB/edit?gid=1472018349#gid=1472018349)
- Change to properties of Library based on GCC feedback [here](https://docs.google.com/spreadsheets/d/1DfnOWWnEZ5xFwbFnF8kGk8kTgWfywqL4swwMINIHvbA/edit?gid=23#gid=23)
- Include more detailed descriptions and examples of some fields that are technology-specific (i.e. Kinnex)
- Minor change to TCC submission schema related to units and changing enums to suggested enums
- Update documentation for setting up Google Sheet API credentials for automated submission templates

Example submission template can be found here:
[submission_workbook.xlsx](https://github.com/user-attachments/files/15842916/submission_workbook.xlsx)


src/encoded/schemas/sequencing.json
```diff
@@ -75,6 +75,12 @@
             "description": "Flow cell used for sequencing",
             "type": "string"
         },
+        "monomer_count": {
+            "title": "Monomer Count",
+            "description": "Monomer count for Kinnex",
+            "type": "integer",
+            "minimum": 0
+        },
```
drio18, line 83: Is this a QC metric we'll be duplicating on our end? If so, consider leaving off. If not, consider moving to the relevant file or file set items rather than here (?); having the field here means slightly different counts will necessitate new sequencing items, which isn't a great use for the item.

src/encoded/tests/test_schema_library.py
```diff
@@ -0,0 +1,36 @@
...
+from webtest.app import TestApp
+from dcicutils.schema_utils import get_property
+from dcicutils.ff_utils import patch_metadata
+
+from .utils import (
+    load_schema,
+    validate_schema,
+    patch_item,
+    get_insert_identifier_for_item_type,
+)
```
drio18, line 13: Many unused imports here to remove.

src/encoded/tests/test_schema_library.py
```diff
@@ -0,0 +1,36 @@
...
+    patch_body: Dict[str, Any],
+    status: int
+)-> None:
+    """Ensure target_monomer_size is only valid if concatenated_reads is Yes."""
+    library_uuid = get_insert_identifier_for_item_type(
+        testapp=es_testapp,
+        item_type="Library"
+    )
+    assert patch_item(es_testapp, patch_body,library_uuid,status=status)
+
```
drio18, line 36: Strongly recommend running python files through a formatter or manually fixing via a linter to ensure consistency (e.g. line length, spaces after commas, removal of white spaces/extra lines, etc.).

src/encoded/schemas/library.json
```diff
@@ -203,6 +156,24 @@
...
+                    }
+                }
+            },
+            "then": {
+                "not": {
+                    "required": [
+                        "target_monomer_size"
+                    ]
+                }
+            }
```
drio18, line 174: Is submission of `concatenated_reads` without `target_monomer_size` allowed?

src/encoded/schemas/analyte.json
```diff
@@ -23,6 +23,26 @@
...
+        "genomic_quality_number_instrument": [
+            "genomic_quality_number"
+        ],
+        "dna_quality_number": [
+            "dna_quality_number_instrument",
+            "dna_quality_size_threshold"
+        ],
+        "dna_quality_number_instrument": [
+            "dna_quality_number"
+        ],
```
drio18, line 45: Strongly consider adding tests for these.

src/encoded/upgrade/library.py
```diff
@@ -12,3 +12,22 @@ def upgrade_library_1_2(
...
+
+@upgrade_step("library", "2", "3")
+def upgrade_library_2_3(
+    value: Dict[str, Any], system: Dict[str, Any]
+) -> Dict[str, Any]:
+    """Delete fragment_maximimum_length, fragment_mean_length, 
+    fragment_minimum_length, fragment_standard_deviation_length,
+    insert_coefficient_of_variation, insert_maximum_length,
+    insert_mean_length, insert_minimum_length, insert_standard_deviation_length,
+    target_insert_maximum_length, target_insert_mean_length, target_insert_maximum_length."""
```
drio18, line 25: Consider removing repetition of property names here in the docstring. Also, check out PEP8 recommendations for docstring formatting.

src/encoded/upgrade/library.py
```diff
@@ -12,3 +12,22 @@ def upgrade_library_1_2(
...
+
+    removed_properties = ["fragment_minimum_length", "fragment_standard_deviation_length",
+    "insert_coefficient_of_variation", "insert_maximum_length",
+    "insert_mean_length", "insert_minimum_length", "insert_standard_deviation_length",
+    "target_insert_maximum_length", "target_insert_mean_length", "target_insert_maximum_length"]
```
drio18, line 30: Consider rewriting this list to make more readable (or run through a formatter):
```suggestion
    removed_properties = [
        "fragment_minimum_length",
        "fragment_standard_deviation_length",
        "insert_coefficient_of_variation",
        "insert_maximum_length",
        "insert_mean_length",
        "insert_minimum_length",
        "insert_standard_deviation_length",
        "target_insert_maximum_length",
        "target_insert_mean_length",
        "target_insert_maximum_length",
      ]
```

src/encoded/schemas/library.json
```diff
@@ -101,83 +101,36 @@
...
-            "title": "Insert Mean Length",
-            "description": "Mean length of the sample molecule in the fragments to be sequenced (bp)",
-            "type": "number",
-            "minimum": 0
-        },
-        "insert_minimum_length": {
-            "title": "Insert Minimum Length",
-            "description": "Minimum length of the sample molecule in the fragments to be sequenced (bp)",
-            "type": "integer",
-            "minimum": 0
```
drio18, line 150: So we don't want QC data on the library fragment sizes anymore? We do have items with these values in prod, and that data won't be easily retrievable once these changes go in.

src/encoded/schemas/library.json
```diff
@@ -101,83 +101,36 @@
...
+            "title": "Target library size",
+            "description": "Targetted size for library (bp)",
+            "type": "integer"
         },
-        "target_insert_mean_length": {
-            "title": "Target Insert Mean Length",
-            "description": "Target mean length of the insert (bp)",
-            "type": "integer",
-            "minimum": 0
-        },
```
drio18, line 175: Shouldn't this value be used on existing items for the new prop above?

CHANGELOG.rst
```diff
@@ -7,6 +7,19 @@ smaht-portal
...
+======
+
+`PR 184: Update submission schema from TPC and GCC feedback <https://github.com/smaht-dac/smaht-portal/pull/184>`_
+
+* Remove SamplePreparation item and merges the property homogenization_method with AnalytePreparation
+* Add properties to Analyte based on GCC Benchmarking Extraction input
+* Changes to properties of Library based on GCC feedback
+* Include more detailed descriptions and examples of some fields that are technology-specific (i.e. Kinnex)
+* Minor changes to TCC submission schema related to units and changing enums to suggested enums
+* Updates documentation for setting up Google Sheet API credentials for automated submission templates
```
drio18, line 20: Consider consistent tense and format here (e.g. like commit messages, start with verb in first person, simple present tense).

drio18 (changes requested):
Primarily dropping stylistic comments, but there are a couple worthy of discussion or changes to me eyes.

src/encoded/schemas/library.json
sarahgonicholson: Do you think it would make sense to add this data in comments for the data that has information for these properties?

drio18: Not really, as I'm not a huge fan of collecting very unstructured data since hard to parse/read and not searchable. If we're removing the properties because we don't think the information is relevant, then the values can be deleted; otherwise, would keep some subset of the properties.

src/encoded/schemas/sequencing.json
sarahgonicholson: I think you're right. Removed this property

src/encoded/tests/test_schema_library.py
sarahgonicholson: I forgot to install a formatter! Done.

src/encoded/schemas/analyte.json
sarahgonicholson: Added tests for dna_integerity, genomic_quality, and dna_quality. I also make all three combinations mutually required. Thanks!

src/encoded/schemas/analyte.json
```diff
@@ -219,6 +337,24 @@
...
+            "title": "Yield per million cells",
+            "description": "Yield per million cells (ug/M cells), if applicable",
+            "type": "number",
+            "minimum": 0
+        },
+        "yield_per_mg_tissue": {
+            "title": "Yield per mg tissue",
+            "description": "Yield per milligram of tissue (ug/mg), if applicable",
+            "type": "number",
+            "minimum": 0
```
drio18, line 357: Not necessary here, as I doubt this data will be searched on, but can always separate out the number from the units. The benefit of the separation approach is that if someone wants to provide different units later on, it's typically a simple addition to the unit enum, whereas the approach above forces either a redesign or more properties to capture essentially identical information. Also, when the field is split between number and unit, search is much more direct.

src/encoded/tests/test_schema_analyte.py
```diff
@@ -31,3 +34,128 @@ def test_rna_integrity_requirements(
...
+    patch_body: Dict[str, Any],
+    expected_status: int,
+    es_testapp: TestApp,
+    workbook: None,
+) -> None:
+    """Ensure mutual requirements for DIN and instrument."""
+    patch_item(
+        es_testapp,
+        patch_body,
+        "3932c9d7-c5e6-46c3-9e67-2ccd276f4b74",
```
drio18, line 64: I know you're following the example above here, but using `get_insert_identifier_for_item_type` as in `test_schema_library.py` is the better approach, since someone may change the UUID here and break all these tests.

src/encoded/schemas/library.json
sarahgonicholson: I think I was a little overzealous with removing all of these. New commit keeps all but four of these properties.

src/encoded/upgrade/library.py
```diff
@@ -12,3 +12,20 @@ def upgrade_library_1_2(
...
+
+    removed_properties = [
+        "fragment_minimum_length",
+        "fragment_maximum_length",
+        "fragment_standard_deviation_length",
+        "insert_standard_deviation_length",
+    ]
+    for property in removed_properties:
+        if property in value:
+            del value[property]
```
drio18, line 31: Good to write a test for this upgrader.

drio18, line 31: Does it make sense to move the existing `target_insert_mean_length` to the new `target_fragment_size`, and then delete the prior property?

src/encoded/schemas/tissue_sample.json
```diff
@@ -100,6 +106,26 @@
...
+            },
+            "then": {
+                "not": {
+                    "required": [
+                        "core_size"
+                    ]
+                }
+            }
+        }
+    ],
```
drio18, line 128: Would be good to add a quick test for this as well. Also, should it be required to have the core size if it is a core?

src/encoded/upgrade/library.py
sarahgonicholson: That probably makes sense. For non-short read libraries, I'm a little unclear on how submitters will be defining fragments vs. inserts

drio18: I think that's generally been confusing and is likely the reason people want this simplified to one property.

src/encoded/schemas/tissue_sample.json
sarahgonicholson: This has actually introduced a larger issue with how we call `get_identifying_insert` in `tests/utils.py` for testing items. Currently the `get_required_properties` function doesn't handle `allOf` statements or `if/then/else` statements for conditionally required properties. I'm working on a work-around that doesn't use either of these statements to address this for now.

sarahgonicholson: I found a workaround using `anyOf` that functions. I've added tests for this and for `target_monomer_size` and `concatenated_reads` in Library.

src/encoded/upgrade/library.py
sarahgonicholson: I've changed the upgrader to convert all `target_insert_mean_length`  values to the `target_fragment_size`.

src/encoded/schemas/analyte.json
sarahgonicholson: I've changed to split fields between number and unit.

CHANGELOG.rst
```diff
@@ -7,6 +7,17 @@ smaht-portal
...
+0.66.0
+======
+
+`PR 184: Update submission schema from TPC and GCC feedback <https://github.com/smaht-dac/smaht-portal/pull/184>`_
+
+* Remove SamplePreparation item and merges the property homogenization_method with AnalytePreparation
+* Add properties to Analyte based on GCC Benchmarking Extraction input
+* Change to properties of Library based on GCC feedback
+* Include more detailed descriptions and examples of some fields that are technology-specific (i.e. Kinnex)
+* Minor change to TCC submission schema related to units and changing enums to suggested enums
```
drio18, line 19:
```suggestion
* Minor change to TPC submission schema related to units and changing enums to suggested enums
```
```suggestion
* Minor change to TCC submission schema related to units and changing enums to suggested enums
```

CHANGELOG.rst
```diff
@@ -7,6 +7,17 @@ smaht-portal
...
+* Update documentation for setting up Google Sheet API credentials for automated submission templates
```
drio18, line 20:
```suggestion
* Update documentation for setting up Google Sheet API credentials for automated submission templates


```

drio18 (approved):
Looks good. You've probably already checked, but ensure you purge any existing SamplePreparation items on existing deployments prior to merging.

wfeng19 (approved):
Everything looks good to me!

PR #192, opened 2024-06-27
fix: include missing TPC and GCC in awardees table

crfmc:
Corresponds to [Trello ticket 301](https://trello.com/c/2cdN64Wc/301-fix-missing-awardees-in-table).
* Small PR to include consortiums into the awardees table.

Deployed to [smaht-devtest](https://devtest.smaht.org/about/consortium/awardees).


alexander-veit (approved):
Looks good.

PR #196, opened 2024-06-28
Bm primary tissue fixes

Bianca-Morris:
### Changelog:
- Hide unexpected facets
- Fix un-synced tab title and table counts (pulling directly from `searchHref` instead of hard-coding... this may need to reference schema in the future depending how much changes)
- Fix for home page link to primary/benchmarking tissue page

### DIFF
--- a/src/encoded/static/components/static-pages/components/benchmarking/BenchmarkingTable.js
+++ b/src/encoded/static/components/static-pages/components/benchmarking/BenchmarkingTable.js
@@ -246,6 +246,8 @@ const BenchmarkingTable = (props) => {
             hideFacets={[
                 'dataset',
                 'file_sets.libraries.analytes.samples.sample_sources.code',
+                'status',
+                'validation-errors',
             ]}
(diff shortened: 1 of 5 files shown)


src/encoded/static/components/static-pages/components/benchmarking/BenchmarkingTable.js
```diff
@@ -243,7 +243,12 @@ const BenchmarkingTable = (props) => {
                 onResetSelectedItems,
             }}
             columnExtensionMap={benchmarkingColExtMap}
-            hideFacets={['dataset']}
+            hideFacets={[
+                'dataset',
+                'file_sets.libraries.analytes.samples.sample_sources.code',
+                'status',
+                'validation-errors',
```
Bianca-Morris, line 250: This doesn't seem to be doing the job. Not sure why, I'll need to look into it more to determine why this facet is still showing up on staging.

crfmc (approved):

PR #200, opened 2024-07-02
SN TPC Schema Update

sarahgonicholson:
Follow-up changes to the TPC submission schema:
- Add `preservation_type` and `preservation_medium` properties to Tissue.
- Make `preservation_type` an enum for Tissue and Sample
- Make `core_size` a string enum
- Add `size_unit` to Tissue to account for tissues measured in cm^2 instead of cm^3


src/encoded/schemas/tissue.json
```diff
@@ -93,6 +101,22 @@
...
+        },
+        "preservation_type": {
+            "title": "Preservation Type",
+            "description": "Method of tissue preservation",
+            "type": "string",
+            "enum": [
+                "Fixed",
+                "Fresh",
+                "Frozen",
+                "Snap Frozen"
```
willronchetti, line 117: These fields are duplicated - please ref the one from sample or add as a mixin

src/encoded/schemas/tissue_sample.json
```diff
@@ -85,8 +85,12 @@
...
             "title": "Core Size",
             "description": "Size of core sample, if applicable (mm)",
-            "type": "number",
-            "minimum": 0
+            "type": "string",
+            "enum": [
+                "1.5",
+                "3.0",
+                "3.0 Donut"
+            ]
```
willronchetti, line 93: Incompatible change - if any values of this currently exists they are incorrectly marked as number and casting may not work - may require an upgrader

willronchetti (commented):
Andy should confirm whether upgrader is needed or if int -> string casting will just work, I'm not totally sure

src/encoded/schemas/tissue_sample.json
aschroed: I think an upgrader would be required, however, there are currently no values for this field in the portal so as long as the change gets in before any data is submitted (and the templates reflect this change it should be OK).

src/encoded/schemas/tissue.json
sarahgonicholson: TPC requested them in Tissue in addition to TissueSample. The thought was that having them in both would allow validation with GCC submissions of TissueSample.
I will ref the one from sample here.

aschroed:
So far I've just addressed Will's comments but will more carefully review when I have a chance.  However, I did notice that while you did include tests, the test inserts for tissue and tissue_sample could use some updates to account for these new and updated properties.  LMK if it is not clear what these are - they are located in: src/encoded/tests/data/inserts

src/encoded/schemas/mixins.json
```diff
@@ -717,6 +717,22 @@
             }
         }
     },
+    "preservation_medium": {
```
aschroed, line 720: These mixins are not quite formatted correctly.  They need a parent container - see 'protocols' below.  These could be grouped under a single parent eg. 'preservation_info' so when you mix it in both properties come along as you are mixing in the same 2 properties in both 'sample' and 'tissue_sample' schemas.

src/encoded/schemas/tissue_sample.json
sarahgonicholson: Yes there are currently no values for this field, as core_size just got added very recently

src/encoded/schemas/tissue.json
```diff
@@ -113,7 +127,7 @@
         },
         "recovery_interval": {
```
aschroed, line 128: If this property is going to remain and a new property of the same name is added to tissue_collection with a different meaning then this could cause confusion.  I don't think this will be a problem for indexing per se as items are namespaced (may want to confirm with @willronchetti ) but general searches could be confounded.  If the goal is to remove this from tissue after the new prop is added to tissue_collection then no worries but consider renaming one or the other?

aschroed (approved):
I left one comment about the time_interval property but otherwise looks fine to me.  Did not run tests.

src/encoded/schemas/tissue.json
sarahgonicholson: Yes, my plan was to remove this property from Tissue once I had transferred the values to a new TissueCollection, unless this people think this is inadvisable.

CHANGELOG.rst
```diff
@@ -101,7 +114,10 @@ Change Log
 * Minor change to TCC submission schema related to units and changing enums to suggested enums
 * Update documentation for setting up Google Sheet API credentials for automated submission templates
 
+<<<<<<< HEAD
+=======
 
+>>>>>>> main
```
willronchetti, line 120: Merge artifacts need to be removed

willronchetti (approved):

PR #203, opened 2024-07-17
Disable controls for non-admins on SubmissionStatus page

alexander-veit:
Disable tagging and commenting functionality for non-admins on Submission Status page


willronchetti (approved):
It should be noted that this is purely a client side change, and non-admins could still change data in the backend - if we're ok with this then that's fine

alexander-veit:
Oh yes, I am aware of that. We mostly want to avoid accidental changes by non-admins through the Submission Status page.

PR #204, opened 2024-07-18
SN TPC schema update

sarahgonicholson:
For MedicalHistory, change `enum` from "Not done" to "Not Done" for consistency


aschroed (approved):
Assuming this property has not been used yet and the values provided in metadata are using the cap form so all looks fine to me.

PR #205, opened 2024-07-18
SN Add arguments to write_submission_spreadsheet for ordered items

sarahgonicholson:
Adds two arguments to `write-submission-spreadsheets` that create submission_workbook templates with a subset of ordered tabs consistent with the submission templates shared with TPCs and GCCs/TDDs.
- `--tpc` outputs tabs for the TPC submission 
- `--gcc` outputs tabs for the GCC/TDD submissions


src/encoded/commands/write_submission_spreadsheets.py
```diff
@@ -850,6 +908,8 @@ def main():
     )
     parser.add_argument("--env", help="Environment", default="data")
     parser.add_argument("--item", help="Item name", nargs="+")
+    parser.add_argument("--tpc", help="TPC Submission items", action="store_true")
+    parser.add_argument("--gcc", help="GCC Submission items", action="store_true")
```
alexander-veit, line 912: They can't both be set, right? --gcc will be ignored then. Might be good to emit a warning if that happens. Not necessary though.

alexander-veit (approved):
Looks good.

PR #206, opened 2024-07-19
SN Remove tissue recovery_interval

sarahgonicholson:
Removes `recovery_interval` property from Tissue schema
- Includes upgrader for tissue schema from version 2 to 3 with test
- All existing `recovery_interval` values for Tissue items have already been transferred to TissueCollection items linked to Donor


aschroed (approved):
Looks good.

PR #210, opened 2024-07-25
SN DSA Schema Update

sarahgonicholson:
Changes to DonorSpecificAssembly and SupplementaryFile schemas to reflect GCC feedback:
- BUSCO score properties (i.e. `percent_single_copy`) changed type to arrays of numbers to reflect multiple haplotypes
- Change property name from `percent_duplicate` to `percent_multi_copy` for clarity
- Change property name from `largest_contig` to `largest_contig_size` for clartiy
- Change enums for `data_type` and `data_category` specific to  SupplementaryFile fasta and chain files
- Fix descriptions of some properties


src/encoded/schemas/donor_specific_assembly.json
```diff
@@ -124,30 +124,42 @@
         "percent_single_copy": {
             "title": "Percent Single Copy",
             "description": "Percentage of highly conserved genes with a single copy present in the assembly (BUSCO)",
-            "type": "number",
-            "minimum": 0,
-            "maximum": 100
+            "type": "array",
+            "items": {
```
aschroed, line 128: I am not familiar with this metric but is it clear to you how it will be used and if there are multiple value is there a need for an explicit ordering or some other way to indicate particular haplotype a number refers to?
Same question for three fields below as well.

aschroed (approved):
I do have one question about the properties that were converted to arrays but other than that looks OK to me.

src/encoded/schemas/donor_specific_assembly.json
sarahgonicholson: Yes this was something I was thinking about. Do you think it might be better to have this be an object with properties of haplotype and value, rather than an array?

aschroed: That is definitely one way to do it - though I know that Doug had wanted to avoid embedded objects like that as much as possible - arrays of embedded objects can be a pain - does submittr and the workbooks have examples of this yet?  May also want to ask @willronchetti what he thinks - it shouldn't be an ES issue but I'm not completely familiar with how SMaHT es is configured for embedded objects.

PR #216, opened 2024-07-30
SN Add ExternalQualityMetric

sarahgonicholson:
- Add `ExternalQualityMetric` item type which is a submitted item inheriting some properties from pre-existing `QualityMetric`
- Create new linking property `external_quality_metrics` in `file.json`
- Add new properties to `variant_calls.json` schema: `comparator`, `external_databases`, and `filtering_methods`


src/encoded/schemas/external_quality_metric.json
```diff
@@ -0,0 +1,106 @@
...
+        },
+        "category": {
+           "items": {
+                "type": "string",
+                "suggested_enum": [
+                    "DuplexSeq"
+                ]
+            }
+        },
+        "qc_values": {
```
willronchetti, line 75: Can you not merge this entire field from quality metric as is without the subfields? If a description is missing it can probably just go on quality metric.

willronchetti (approved):

src/encoded/schemas/external_quality_metric.json
sarahgonicholson: I did it this way because the other properties within `qc_values`, `visible` and `flag` seemed to be things that were specific to internal QualityMetric items. Would you recommend merging the whole property and setting those to `restricted_fields` or something else?

alexander-veit (approved):

PR #225, opened 2024-08-12
SN Improvements to automated submission spreadsheet

sarahgonicholson:
In `write-submission-spreadsheet`:
- clean up args messages
- Add functionality for `--item` working with `--google`
- Adjust comment descriptions to clarify | delimiter and add search links for submitted items

In `schemas/`:
- Include schema `description` for GCC-submitted items. Descriptions can also be found [here ](https://docs.google.com/document/d/1uJ18-hhO1GjRsMqzCrFVYXLJzsg6RaUUmIkav3pSGks/edit)


aschroed (approved):

src/encoded/schemas/file_set.json
```diff
@@ -1,5 +1,6 @@
 {
     "title": "File Set",
+    "description": "A FileSet item consists of the file produced from sequencing a library. Note: each library generated requires its own unique associated FileSet.",
```
wfeng19, line 3: I think "file" in "A FileSet item consists of the file produced..." should be plural.

src/encoded/schemas/treatment.json
```diff
@@ -1,5 +1,6 @@
 {
     "title": "Treatment",
+    "description": "A non-standard treatment applied during analyte preparation or library preparation.",
```
wfeng19, line 3: What does "non-standard" mean in this context? Is the implication that "standard" treatments would fall under PreparationKit?

src/encoded/schemas/supplementary_file.json
```diff
@@ -1,5 +1,6 @@
 {
     "title": "Supplementary File",
+    "description": "Submitted files not generated by DAC",
```
wfeng19, line 3: I'm not sure if this description is accurate, since we have DAC-generated files that will be submitted as supplementary files (the downsampled HG005 files from Alice Lee's lab) and not all non-DAC-generated submitted files are considered supplementary files. Maybe we can say that supplementary files are files that do not fall into other file item types in the schema, namely UnalignedReads, AlignedReads, and VariantCalls? I'm not sure if this is technically true either, though; there's probably a better description for SupplementaryFiles than this.

src/encoded/schemas/treatment.json
sarahgonicholson: That was the intention, although I am still having a hard time coming up with examples of a Treatment. Maybe "A treatment not from a kit"?

src/encoded/schemas/supplementary_file.json
sarahgonicholson: That's a good point. Maybe SupplementaryFiles are "not generated by the DAC analysis pipeline"? Or should I just say that they are files that don't fit into any of the other file types?

wfeng19: I think what you changed the description to is more accurate! I am not sure if it's 100% accurate still, since we generally don't process any of the TTD files as well, but it should be alright for now.

PR #230, opened 2024-08-16
SN Update annotated_filenames

sarahgonicholson:
In `commands/create_annotated_filenames.py`:
- For annotated filenames, update `aliquot_id` to be `[aliquot_id]MC` if multiple tissue samples from the same tissue aliquot but multiple cores (e.g. ST001-1A-001A1 and ST001-1A-001B2) and `MAMC` if from multiple tissue samples from different tissue aliquots
- Remove the variant type from the end of annotated filenames for vcfs

In `commands/write_submission_spreadsheets.py`
- Add notes for properties that allow commas for large numbers and allow abbreviations (e.g. 100M or 3.1 Gb)

In `schemas/file.json`
- Add a `version` property for front-end


src/encoded/commands/create_annotated_filenames.py
```diff
@@ -654,11 +650,29 @@ def get_aliquot_id_from_samples(tissue_samples: List[Dict[str], Any]) -> Filenam
...
         get_aliquot_id_from_tissue_sample(sample) for sample in tissue_samples
     ]
+    if len(aliquot_ids) > 1:
+        aliquot_ids = get_multiple_aliquot_id_from_samples(aliquot_ids)
     return get_filename_part_for_values(
         aliquot_ids, "tissue aliquot ID", source_name="sample"
     )
 
 
+def get_multiple_aliquot_id_from_samples(ids: List[str]):
```
aschroed, line 660: As I am not clear on how the decision on this ended up and the desired result I leave it up to you to make sure that this is doing as expected.

aschroed (approved):
Looks fine to me.

src/encoded/commands/create_annotated_filenames.py
```diff
@@ -654,11 +650,29 @@ def get_aliquot_id_from_samples(tissue_samples: List[Dict[str], Any]) -> Filenam
...
+    """Get filename part for files merged from multiple tissue sample aliquots."""
+    aliquot_ids = []
+    core_ids = []
+    for id in ids:
+        aliquot_id = id[:-2]
+        core_id = id[-2:]
```
willronchetti, line 666: You have an implicit invariant here that must be true in order for calls like this to succeed - so long as there is external validation elsewhere that ensures this, it's probably fine, but may merit a comment to illustrate what you are extracting.

src/encoded/schemas/file.json
```diff
@@ -207,6 +210,10 @@
                 "title": "Notes to tsv file",
                 "type": "string"
             }
+        },
+        "version": {
+            "type": "string",
+            "permission": "restricted_fields"
```
willronchetti, line 216: Intended UX for this? May merit some further discussion on how we want to manage "versions" of files

willronchetti (approved):
Needs version bump/CHANGELOG

src/encoded/schemas/file.json
sarahgonicholson: @willronchetti This is intended for the COLO829 Challenge data/Truth set, where presumably there might be multiple challenges and we want to keep track of which one the files came from. Open to suggestions on other ways to handle this though!

PR #232, opened 2024-08-23
SN RE property

sarahgonicholson:
Add a property `restriction_enzymes` to LibraryPreparation
Add columns to SubmissionCenter search page for `identifier` and `code`
Add `description` property to VariantCalls


willronchetti (approved):
Needs version bump + CHANGELOG

src/encoded/schemas/library_preparation.json
```diff
@@ -125,6 +125,21 @@
                 ]
             }
         },
+        "restriction_enzymes": {
```
aschroed, line 128: I suspect string values are fine here.  in 4DN we have enzyme items as it was deemed important for tracking enzyme source and I believe info about the recognition sites are used - perhaps indirectly - in the processing pipeline for Hi-C.  Just something to consider and maybe ask bioinformatics team about in case they might use info about the particular enzyme used in processing?

aschroed (approved):
Looks good.  Added a non-blocking comment on the enzyme property.

src/encoded/schemas/library_preparation.json
sarahgonicholson: That's a good point. I will ask!

PR #241, opened 2024-09-04
SN Add properties to library

sarahgonicholson:
* Add properties `target_dna_sequence`, `guide_rna_sequence`, and `antibody` to `Library` schema
* Change property `restriction_enzymes` to `enzymes`  in `LibraryPreparation` (recently added property so it is empty in the portal) to apply for transposase enzymes as well and update associated tests.

### DIFF
--- a/src/encoded/schemas/library.json
+++ b/src/encoded/schemas/library.json
@@ -98,7 +98,11 @@
         "antibody": {
             "title": "Antibody",
             "description":"Antibody used for affinity enrichment",
-            "type": "string"
+            "type": "string",
+            "suggested_enum": [
+                "H3K4me3",
+                "RNAPII"
+            ]
         },
         "barcode_sequences": {
             "title": "Barcode Sequences",
--- a/src/encoded/schemas/library.json
+++ b/src/encoded/schemas/library.json
@@ -130,10 +130,15 @@
             "type": "number",
             "minimum": 0
         },
-        "guide_rna_sequence": {
-            "title": "Guide RNA Sequence",
-            "description": "Guide RNA sequence. Relevant for Cas9 Nanopore",
-            "type": "string"
+        "guide_sequence": {
+            "title": "Guide Sequence",
+            "description": "Guide sequence. Relevant for Cas9 capture experiments",
+            "type": "array",
+            "uniqueItems": true,
+            "minItems": 1,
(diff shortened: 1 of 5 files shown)


src/encoded/schemas/library.json
```diff
@@ -95,6 +95,11 @@
             "type": "number",
             "minimum": 0
         },
+        "antibody": {
```
aschroed, line 98: May want to add some enum or suggested enum here or you are likely to get a wide variety of values input.  In 4DN we have an antibody item and the 'title' is a calcprop that combines target and vendor.  If an Item is not needed (it likely isn't) then you may want to set up some rules for naming conventions.

aschroed (approved):
I left a couple of comments for your consideration but non-blocking.

src/encoded/schemas/library.json
```diff
@@ -125,6 +130,11 @@
             "type": "number",
             "minimum": 0
         },
+        "guide_rna_sequence": {
+            "title": "Guide RNA Sequence",
+            "description": "Guide RNA sequence. Relevant for Cas9 Nanopore",
```
wfeng19, line 135: Maybe change description to "Relevant for Cas9 capture experiments"?

src/encoded/schemas/library.json
```diff
@@ -155,6 +165,11 @@
             "type": "string",
             "format": "date"
         },
+        "target_dna_sequence": {
+            "title": "Target DNA Sequence",
+            "description": "Target DNA sequence used in insert selection",
```
wfeng19, line 170: Was this property also intended for the Cas9 capture experiments? It looks like we added a "dna_target" property on the UMich metadata sheet (https://docs.google.com/spreadsheets/d/1sFOEEc_JtE-wNO7RnN4EcaEVeBuDU8uMNHnAJzoLc2s/edit?gid=493168297#gid=493168297), but I wasn't sure if this property was intended to be used in the same way.

wfeng19 (commented):
I left a couple of comments, but nothing non-blocking as well.

src/encoded/schemas/library.json
sarahgonicholson: I see, I wasn't sure exactly what the intended values were for this. Updating!

PR #244, opened 2024-09-06
SN FileSet calcprop

sarahgonicholson:
- Create calcprop for FileSet, `files_status_retracted`, that returns "True" if a file in `files` has the status of `obsolete` or `retracted`
- Embed `file_set.files_status_retracted` in File


Bianca-Morris:
Just pulled again and tested this locally - seems to be working as it should! Let me know if you need another look.

Bianca-Morris (approved):

src/encoded/tests/test_types_file_set.py
```diff
@@ -123,3 +127,12 @@ def test_validate_compatible_assay_and_sequencer_on_post(
...
 
+
+@pytest.mark.workbook
+def test_associated_files_retracted(es_testapp: TestApp, workbook: None) -> None:
+    """Ensure `associated_files_retracted` calc_prop is working."""
+    status_search = get_search(
+        es_testapp,
+        "?type=FileSet&files_status_retracted=True",
+    )
+    assert status_search
```
willronchetti, line 138: I would do a count check here at least to ensure you're getting the right search results

willronchetti (approved):

dmichaels-harvard (approved):

PR #250, opened 2024-09-11
Server-side validation skip when skip_links=True

dmichaels-harvard:
Effectively disable server-side validators which reference linked objects for smaht-submitr, when the skip_links=True.


willronchetti (approved):
Worth testing with a dummy collection if possible

sarahgonicholson (approved):

PR #271, opened 2024-10-08
SN Links to Existing Data

sarahgonicholson:
- Update links and text descriptions for Links to Existing Data page


docs/source/referencing_data.rst
```diff
@@ -2,21 +2,48 @@
...
+
+.. image:: /static/img/docs/data_model.png
+  :target: /static/img/docs/data_model.png
+  :alt: Data Model Image
+
+
+This allows for clear interpretation of the origin of sequence data.
 References to existing data are frequently required to submit new data. The links below will display all options available for referencing data already in the database.
 
+In the submission spreadsheet (See <a href="/docs/submission/getting-started-with-submissions", target="_blank">Getting Started with Submissions</a>), columns that are italicized indicate fields that should be unique identifiers for items of another type.
```
Bianca-Morris, line 17: Ah I see the issue. So there are a few different ways to handle in-line links in RST. You could do it inline like this (like the syntax used in the changelog):
```
In the submission spreadsheet (See `Getting Started with Submissions <docs/submission/getting-started-with-submissions>`_), columns that are italicized indicate fields that should be unique identifiers for items of another type.
```
or via reference:
```
.. _getting-started-with-submissions: docs/submission/getting-started-with-submissions

In the submission spreadsheet (See `Getting Started with Submissions`_), columns that are italicized indicate fields that should be unique identifiers for items of another type.
```

Unfortunately, RST doesn't have an equivalent to the `target="_blank"` functionality from HTML. So if you want to open the link in a new tab, you'll need to make that paragraph an HTML paragraph instead. That would look something like this:
```
.. raw:: html

<p>
In the submission spreadsheet (See <a href="/docs/submission/getting-started-with-submissions", target="_blank">Getting Started with Submissions</a>), columns that are italicized indicate fields that should be unique identifiers for items of another type.
</p>
```

Bianca-Morris (commented):
I pushed a small commit to add a bit of vertical padding to the data model image. If you are OK with doing more tweaks on that image, you might reach out to Shannon, who can help suggest colors and fonts. Other than that, I think this is fine. At some screen sizes the images aren't full width, which is something you could fix by adding some additional horizontal padding, but I don't think it's 100% necessary.

See my note on the links.

Bianca-Morris:
[Improve Links to Existing Data documentation](https://trello.com/c/56UHd66j/337-improve-links-to-existing-data-documentation)

Bianca-Morris (approved):
I think I ordered some things in the referencing data page slightly differently, so those conflicts will need to be addressed, but this looks good to merge to me.

PR #287, opened 2024-11-07
SN Override Coverage

sarahgonicholson:
Add property to file, `override_group_coverage` that displays in `data_generation_summary` calculated property for the File Overview page, which takes precedence  over `file_set.sequencing.target_coverage` if set


aschroed (approved):
Looks good to me!

dmichaels-harvard (approved):

PR #295, opened 2024-11-20
Adding /browse view.

dmichaels-harvard:


willronchetti (approved):

Bianca-Morris (commented):
This is a good starting point, but it currently only seems to work for File items? We need support for Donors, Assays, and Tissues, as well. I can start working on the layout with this for now, though, so if you want to merge this and add that additional functionality later, that should be OK.

PR #298, opened 2024-11-21
SN File release tracker

sarahgonicholson:
* Add calcprop to file returning concatenated string of `file_sets.libraries.assays.display_title`, `file_sets.sequencing.sequencer.display_title`, and `file_format.display_title` (e.g. "Fiber-seq PacBio Revio BAM", "Illumina WGS BAM")
* If there are multiple values for assay or sequencer, return an empty string

### DIFF
--- a/src/encoded/types/file.py
+++ b/src/encoded/types/file.py
@@ -712,7 +712,7 @@ class File(Item, CoreFile):
             request_handler = RequestHandler(request=request)
             result = self._get_release_tracker_description(
                 request_handler,
-                file_properties=self.propertes
+                file_properties=self.properties
             )
         return result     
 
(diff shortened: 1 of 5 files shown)


src/encoded/types/file.py
```diff
@@ -979,6 +1000,39 @@ def _get_analysis_summary_fields(
...
+            file_utils.get_sequencers(file_properties, request_handler)),
+            item_utils.get_display_title,
+            )
+        file_format_title = get_property_value_from_identifier(
+                request_handler,
+                file_utils.get_file_format(file_properties),
+                item_utils.get_display_title,
+            )
+        if len(assay_title) > 1 or len(sequencer_title) > 1:
+            # More than one unique assay or sequencer
```
dmichaels-harvard, line 1025: Not sure it's a real case, but *could*, if multiple assays and/or sequencers, then if the same number of both, then just return the first one (?) - not sure that makes sense - just thinking ...

sarahgonicholson: @dmichaels-harvard I think the only instance where we'd expect there to be multiple assays or sequencers is for files like DSA or Truth Set files where multiple datasets were used to generate the file. For those special cases, do you think it is work including an override property to allow us to manually set this value?

dmichaels-harvard (approved):

PR #304, opened 2024-12-06
SN remove recovery_datetime from tissue

sarahgonicholson:
- Remove `recovery_datetime` from Tissue after having moved the property to TissueCollection, as it is protected information


aschroed (approved):
Looks good to me.

PR #311, opened 2024-12-20
SN Move donor properties

sarahgonicholson:
- For protection of potentially identifiable information, move properties `height`, `weight`, and `body_mass_index` to MedicalHistory. These will later be removed from Donor
- Move `hardy_scale` to Donor, as this can be public information. Later will be removed from DeathCircumstances


aschroed (approved):
Looks fine to me.

PR #316, opened 2025-01-20
Show status of unaligned reads on Submission Status page

alexander-veit:
This PR adds a badge displaying `archived` or `deleted` to the Submission column on the Submission Status page. That makes it easier track submitted files that need to be archived.


crfmc (approved):
Looks good to me

PR #319, opened 2025-01-23
SN Remove properties from donor

sarahgonicholson:
- Remove `height`, `weight`, and `body_mass_index` from Donor, as these properties have been transferred to MedicalHistory
- Remove `hardy_scale` from DeathCircumstances, as this property has been transferred to Donor


aschroed (approved):
Stating the obvious which i'm sure you have already confirmed - make sure none of the existing items have these properties and you likely already did this when you added the properties to their new locations but ensure that test workbooks/inserts/fixtures have been updated.

sarahgonicholson:
Yes the existing items with this metadata have been transferred to the new item on all environment and in test inserts!

PR #323, opened 2025-01-24
SN Ontology terms

sarahgonicholson:
- Update properties for OntologyTerm item to be implemented with Tissue `uberon_id`
- Add parent item Ontology


aschroed (commented):
If the idea is to keep it simple as possible right now then my only suggestion would be to use 'slim_term' rather than 'parent_term' or alternatively rename the property to something like 'grouping_term' (however in that case some potentially re-usable code from 4DN may need to be updated to work the same way).  The term you want to group by is not always a direct parent and may be a couple of levels up the ontology hierarchy.

If you did want to plan for the potential automation of updates as well as leverage some of the calcprops and validators that exist in 4DN (for example look at biosource.py) you might want to transfer the slim_terms and is_slim_for properties from fourfront.

Something else to consider is that is do you want to rely on the prefix in the identifier to specify the ontology.  It is likely fine for this use case, however, many application ontologies including UBERON import terms from other ontologies and the term ids or identifiers start with prefixes other than UBERON.  Not sure that this would be an issue but definitely something to consider.

Let me know if you have questions or want to discuss.

sarahgonicholson:
@aschroed That makes sense, I'll change the `parent_term` to better reflect this. For your second point, do you think having a property like `ontology` that specifies that a term is UBERON, or some other ontology would make sense? Thanks for your comments!

aschroed:
@sarahgonicholson I think a string property for now for the ontology name would be fine for now and may be all you ever need.

sarahgonicholson:
@aschroed I ended up adding an Ontology item, as I think on principal it would be better to have, and may prove to be helpful later on

aschroed (approved):

PR #332, opened 2025-01-31
feat: add animation to sidebar

crfmc:
- Add animation to sliding sidebar


Bianca-Morris (approved):
Looks great! Thanks for addressing this.

PR #334, opened 2025-02-03
feat: tooltip for sample information definitions

crfmc:
[Trello Ticket 396](https://trello.com/c/iK9ARDsu/396-add-tooltip-for-specimen-and-homogenate-definitions)
- Adds tooltips for specific values in sample summary


utku-ozturk (approved):
looks good to me! my only concern is do we need snake-case (`title_popover`) instead of camel case?

PR #338, opened 2025-02-06
Bm browse view2

Bianca-Morris:
**Changelog:**
- Various fixes:
- Add link to browse view from production data arrow
- Enlarge size of file data modal image
- Rework statistics data components to use new search method


crfmc (approved):
Looks good to me. I think we also wanted to add a bit of space between the value and units, for example between the "0" and "KB". If you are making that change elsewhere then never mind!

Screenshot 2025-02-10 at 1 11 44 PM

PR #345, opened 2025-02-14
SN TPC tissue sample validator

sarahgonicholson:
- Add custom validator to TissueSample that checks if GCC-submitted items match certain metadata from TPC-submitted items already present in the portal


aschroed (approved):
I think this looks OK.  There are a couple of unused variables declared but not a huge deal.  And I did not test directly but code looks like it should do what is intended.

dmichaels-harvard (approved):

PR #349, opened 2025-02-18
SN OntologyTerm update

sarahgonicholson:
- Change `uberon_id` in Tissue from a string to a linkTo to OntologyTerm
- Add upgrader for tissue
- Update File `sample_summary` to include tissue information from OntologyTerm

image


src/encoded/item_utils/ontology_term.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Get top grouping term associated with ontology term recursively.
+    
+    If grouping_term is not present, return display_title of item.
+    """
+    grouping_term = get_grouping_term(properties)
+    if grouping_term:
+        to_get = set(
+            get_property_values_from_identifiers(
+                request_handler, [grouping_term], item_utils.get_at_id
```
aschroed, line 31: Are you switching from get_at_id to get_uuid (and using uuid var names) on purpose?

sarahgonicholson: No, that was not on purpose, good catch! Fixing

PR #350, opened 2025-02-19
SN Override Release Tracker

sarahgonicholson:
- Add property `override_release_tracker_description` to file.json that can set the calcprop `release_tracker_description` to desired value
- Embed `donor_specific_assembly.donors` and `donor_specific_assembly.cell_lines.code` in file.py
- In `release-file.py`, require that a file has `release_tracker_description` set prior to release


src/encoded/tests/test_types_file.py
```diff
@@ -1225,27 +1225,24 @@ def test_release_tracker_description(es_testapp: TestApp, workbook: None) -> Non
...
+            file_utils.get_assays(file)[0]
+        )
+        sequencer_from_calcprop = item_utils.get_display_title(
+            sequencing_utils.get_sequencer(
+                file_utils.get_sequencings(file)[0]
+            )
+        )                 
+        description_from_calcprops=f"{assay_from_calcprop} {sequencer_from_calcprop} {file_format}"
+    if "override_release_tracker_description" in file:
+        description_from_calcprops=f"{file_utils.get_override_release_tracker_description(file)} {file_format}"
```
dmichaels-harvard, line 1243: Do you think a possible fallback for this value, if it doesn't exist from where you're pulling (or maybe it always will?) is from the "description" property, which I see from at least a couple examples is populated with (at least it seems to me) a reasonable value (e.g. for c24f7bdd-8ea1-4a63-9e6d-c15144353299 description is: Zipped VCF with germline variant sites in COLO829-BL).

dmichaels-harvard (approved):

src/encoded/tests/test_types_file.py
sarahgonicholson: I think for OutputFile `description` would be an okay fallback, but for user-submitted files, `description` is highly variable and not always filled out, so I don't know how well that would work. I adjusted the `release-file` script so that if ensures that `release_tracker_description` has a value in order to release a file, so I think that should work to remind us to have the override property set if its necessary

PR #361, opened 2025-02-27
fix: comment out unused Submissions page

crfmc:
[Trello Ticket 638](https://trello.com/c/gHWAWTti/638-fix-submissions-page-does-not-exist-remove?filter=member:crfmc)
- Comment out the submissions page calculated property


utku-ozturk (changes requested):
Not sure but the code blocks below also might be commented out:

https://github.com/smaht-dac/smaht-portal/blob/f45d25a9cc3a85cd487714f70d1a005cf4ef7460/src/encoded/static/components/navigation/components/AccountNav.js#L164-L166

https://github.com/smaht-dac/smaht-portal/blob/f45d25a9cc3a85cd487714f70d1a005cf4ef7460/src/encoded/static/components/app.js#L1396-L1401

utku-ozturk (approved):

PR #370, opened 2025-03-07
Add liquid category check

sarahgonicholson:
- Add check to TissueSample ensuring that items with external_id values with protocol id for blood and buccal swab (3A and 3B) have `category` "Liquid" and  fibroblasts (3AC) have `category` of "Cells"


aschroed (approved):
Looks good.

PR #371, opened 2025-03-07
Add uberon_id check for tissue

sarahgonicholson:
- Add property `valid_protocol_ids` to OntologyTerm that is a list of protocol IDs (e.g. 1D, 3A), the portion of `external_id` indicating tissue types defined [here](https://staging.smaht.org/docs/additional-resources/sample-file-nomenclature)
- Add a check to Tissue ensuring that the protocol ID in `external_id` is among the `valid_protocol_ids` for the OntologyTerm linked with `uberon_id`
- Add checks to Donor and Tissue ensuring that `external_id` matches expected format for Benchmarking and Production
- Makes `external_id` a required property for Donor

### DIFF
--- a/src/encoded/item_utils/ontology_term.py
+++ b/src/encoded/item_utils/ontology_term.py
@@ -57,4 +57,4 @@ def get_top_grouping_term(
 
 def get_valid_protocol_ids(properties: Dict[str, Any]) -> str:
     """Get valid_protocol_ids from properties."""
-    return properties.get("valid_protocol_ids","")
+    return properties.get("valid_protocol_ids",[])
--- a/src/encoded/item_utils/ontology_term.py
+++ b/src/encoded/item_utils/ontology_term.py
@@ -53,3 +53,8 @@ def get_top_grouping_term(
         )
     else:
         return item_utils.get_display_title(properties)
+    
+
+def get_valid_protocol_ids(properties: Dict[str, Any]) -> str:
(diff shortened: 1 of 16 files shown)


src/encoded/item_utils/ontology_term.py
```diff
@@ -52,4 +52,9 @@ def get_top_grouping_term(
...
         )
     else:
-        return item_utils.get_display_title(properties)
\ No newline at end of file
+        return item_utils.get_display_title(properties)
+    
+
+def get_valid_protocol_ids(properties: Dict[str, Any]) -> str:
+    """Get valid_protocol_ids from properties."""
+    return properties.get("valid_protocol_ids","")
```
aschroed, line 60: Shouldn't the default get result be an empty array?

aschroed (approved):
I think this looks fine though I wonder how controlled the external_id is.  It appears to assume a hyphen separating 2 'meaningful' values?  From what I can tell the schema nor tissue validators check that this is the case?  Maybe it's there and I missed it as things are so spread around in utils etc. Probably not a big deal but seems like there could be possibility of exceptions being thrown by some of the utils functions if for example there is not a hyphen.

sarahgonicholson:
@aschroed Thanks for your comments! That is a great point. I added checks to Tissue and Donor to make sure that `external_id` is in the expected format for Benchmarking and Production tissues

PR #373, opened 2025-03-11
SN Add property replaced_by

sarahgonicholson:
- Add property `replaced_by` to File to link to replacement files for files that are retracted or obsolete


aschroed (approved):
Looks fine.

PR #375, opened 2025-03-12
SN gpu upgrade

sarahgonicholson:
- Change the property `gpu` to `gpu_architecture` in Software
- Include upgrader with test

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@ smaht-portal
 Change Log
 ----------
 
+0.150.0
+=======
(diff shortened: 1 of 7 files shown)


aschroed (approved):
This looks fine but is it really necessary?  Can't you just change the title?  Maybe not I guess if you list the actual property names in the workbook - OK never mind - probably best to make property names as clear as possible.

sarahgonicholson:
Yeah I think we'd just had some confusion from submitters on what exactly we were asking for here, and thought this might clarify it a bit.

PR #376, opened 2025-03-12
SN search columns facets

sarahgonicholson:
- Add `columns` and `facets` to items frequently used for searching during submission


aschroed (approved):
Looks good to me.

PR #378, opened 2025-03-14
SN Move ONT validator

sarahgonicholson:
- Move custom validator for basecalling software for ONT files to UnalignedReads from SubmittedFile, as this does not need to be applied to every file item type, just raw sequencing reads


aschroed (approved):
Looks good

PR #385, opened 2025-03-27
SN Fix release tracker calcprop

sarahgonicholson:
- Fix the `release_tracker_title` and `release_tracker_description` calcprops  to have the override property function in the case of having multiple values from `file_sets`
- Make `file_sets` a required property for SupplementaryFile, as it is for all other SubmittedFile types, so that the File Overview display and Manifest File generation work properly

### DIFF
--- a/src/encoded/types/file.py
+++ b/src/encoded/types/file.py
@@ -946,7 +946,7 @@ class File(Item, CoreFile):
         return {
             key: value for key, value in to_include.items() if value
         }
-    
+
     def _get_group_coverage(
         self, request_handler: Request, file_properties: Optional[List[str]] = None
     ) -> Union[List[str], None]:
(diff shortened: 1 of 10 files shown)


src/encoded/types/file.py
```diff
@@ -1081,26 +1081,19 @@ def _get_release_tracker_title(
...
-                    return None
-                to_include = item_utils.get_code(cell_line_title[0])
+                to_include = None if len(cell_line_title) > 1 else item_utils.get_code(cell_line_title[0])
             elif (tissue_title := request_handler.get_items(
                 file_utils.get_tissues(file_properties, request_handler)
             )):
-                if len(tissue_title) > 1:
-                    return None
-                to_include = item_utils.get_display_title(tissue_title[0])   
+                to_include - None if len(tissue_title) > 1 else item_utils.get_display_title(tissue_title[0])
```
aschroed, line 1092: Is that a typo?  - -> =?

sarahgonicholson: Oops! Thank you

PR #387, opened 2025-03-28
QC metrics page improvements

alexander-veit:
- Redesign of the Key metrics tab
- Added Tissue classification table for RNA-Seq data under Key metrics tab
- Add "Metrics by file" tab, that can be linked from the file overview page
- Alerts on the sample contamination heatmap are now determined on a file (and not donor) level. Retracted files do not generate a warning.

TODO: version bump


src/encoded/static/components/viz/QualityMetricVisualizations/SampleContaminationHeatmap.js
```diff
@@ -0,0 +1,323 @@
...
+    };
+
+    useEffect(() => {
+        let doRerender = false;
+        if (oldRerenderNumber.current !== rerenderNumber) {
+            oldRerenderNumber.current = rerenderNumber;
+            doRerender = true;
+        }
+
+        if (!isDrawn.current || doRerender) {
```
crfmc, line 66: I was looking into this `isDrawn` ref we've been using to prevent re-renders in the visualizations. One other approach, maybe slightly cleaner, is to:
1. Pass the `selectedDonor` state variable down from the parent component
2. Add `selectedDonor` to the dependency array for the useEffect (to run useEffect when new donor selected)
3. Replace `isDrawn.current = true` in the cleanup function with `isDrawn.current = selectedDonor`

That way, we don't need the `doRerender`, and we can just check:
`selectedDonor !== isDrawn.current // only redraw if selectedDonor has changed`

alexander-veit: Thank you for this suggestion. I will make a note of that and try that out in a future version.

PR #388, opened 2025-03-28
SN Add retraction_reason

sarahgonicholson:
- Add property `retraction_reason` to File for display in the [Data Retraction](https://devtest.smaht.org/resources/retracted-files) page and File Manifest


utku-ozturk (approved):
Looks good to me!

PR #394, opened 2025-04-03
Update sequencing validator to only enfore RNA properties

sarahgonicholson:
- Update the Sequencing validator on FileSet so that it only enforces for RNA and not DNA


aschroed (approved):
Maybe add a test for the force_pass option if one doesn't already exist?
Otherwise looks OK to me.

sarahgonicholson:
@aschroed Good idea, will do!

PR #398, opened 2025-04-08
SN Create resource file

sarahgonicholson:
- Create new file item type, ResourceFile, for DAC-generated files not used in analysis pipelines


aschroed (approved):
I think this looks OK.

PR #407, opened 2025-04-23
SN upgradeAnalytePrep 

sarahgonicholson:
- Add upgrader changing property `cell_sorting_method` to `cell_selection_method` in AnalytePreparation and update enums
- Add enums to File `data_category` and `data_type` for STORM-Seq metadata files


aschroed (approved):
Looks good.

PR #410, opened 2025-04-28
fix: use retraction_reason in file view alert

crfmc:
* Use retraction_reason property to create file overview alert string instead of notes_to_tsv


crfmc:
[Use retraction_reason for alert banner in file overview](https://trello.com/c/Ec1gVDq0/660-use-retractionreason-for-alert-banner-in-file-overview)

utku-ozturk (approved):

PR #413, opened 2025-04-29
SN Generic Config item

sarahgonicholson:
- Add new item, GenericConfig, that can be used to hold JSON objects


aschroed (approved):
I think this looks OK.

alexander-veit (approved):

willronchetti (approved):

PR #419, opened 2025-05-07
Update MWFR inputs calcprop

alexander-veit:
* Only show MWFRs that are not deleted on the file
* Adjust the QC metrics page JSON script to use the new `tissue_types` property on `Fileset`


sarahgonicholson (approved):
Looks good!

PR #422, opened 2025-05-08
SN Update treatment

sarahgonicholson:
- Add `description` property to Treatment
- Add `dependentRequired` for `concentration` and `concentration_units`


sarahgonicholson:
@aschroed Really great points and I'll probably need to think about this more. What prompted this change was a submitter said that the agent they’re using for DNA extraction is actually a mix of phenol, chloroform, and isoamyl alcohol and they have this in a ratio of 25:24:01. I'm not sure if they would want to include the `concentration` values for each agent individually (but future submitters might). So it may also be work making `concentration` and `concentration_units` arrays as well.

For your second question, I think that case above is multiple agents in a treatment as they used a mixture, but if there were multiple treatments per preparation we would just ask them to make separate Treatment items.

aschroed:
@sarahgonicholson depending on order for multiple associated arrays of items can be a pain.  In the phenol/chloroform/isoamyl case you cite that is a pretty standard (or at least used to be) reagent mix and it also seems somewhat unlikely that people would want to query each agent separately?  I might consider having this mixture as a single string with the concentration info the same.  That said there may be some relevance in terms of effect of this reagent on base calling or sequencing and if so I may revise my comment.

sarahgonicholson:
@aschroed I agree I think it makes sense to just take in the `agent` as a single string with a single concentration. I switched back `agent`

PR #429, opened 2025-05-16
QC metrics page improvements

alexander-veit:
- Improved warnings
- Adjusted colors for tags
- Fixed bug were contents of "Metrics By File" tab contents would not be correctly displayed when preselected tab was not "Metrics By File"
- Added grouping by donor


crfmc (commented):
Looks good!

PR #430, opened 2025-05-19
Cypress Tests for Experimental Assay Terms Grouping

utku-ozturk:
- Updates for `06_benchmarking` and `04a_search_views_local` upon updates in facet terms grouping
- Add new steps into `03a_browse_view_basic` for facet terms grouping for the Experimental Assay facet for both Include Properties and Exclude Properties


crfmc (approved):
Looks good to me!

PR #439, opened 2025-05-29
SN Ontology term germ layer

sarahgonicholson:
- Change the ontology_term function `get_top_grouping_term` to `get_grouping_term_from_tag`, which will recursively search links of OntologyTerm items by `grouping_term` to find the item with tags that contain `tag`
- Add calculated property `category` to Tissue which uses OntologyTerm to classify tissue as a germ layer (Mesoderm, Ectoderm, Endoderm), Germ Cells, or Clinically Accessible
- Add `category` to the `sample_summary` calculated property to be accessible from file
image


aschroed (approved):
I think this looks OK. 
I don't know how this is expected to look but what happens if category is None?

sarahgonicholson:
> I think this looks OK. I don't know how this is expected to look but what happens if category is None?

If category is None, which would only be for non-SMaHt tissues from TTDs, it will just be empty in the file and tissue calc props, which I believe is okay

PR #442, opened 2025-06-02
SN Add coverage calc prop

sarahgonicholson:
- Add `coverage` calculated property to QualityMetric and embed it on File for use in the Data Matrix

### DIFF
--- a/src/encoded/types/quality_metric.py
+++ b/src/encoded/types/quality_metric.py
@@ -1,3 +1,4 @@
+from typing import List, Dict, Any
 from pyramid.view import view_config
 from snovault import collection, load_schema, calculated_property
 from encoded_core.qc_views import download as qc_download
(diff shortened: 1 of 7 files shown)


src/encoded/types/quality_metric.py
```diff
@@ -28,6 +30,19 @@ class QualityMetric(Item):
...
+    @calculated_property(
+        schema={
+            "title": "Coverage",
+            "type": "number",
+            "description": "Estimated average coverage from QC values",
+        }
+    )
+    def coverage(self, request, qc_values: List[Dict[str, Any]]) -> str:
+        for qc in qc_values:
+            if "derived_from" in qc:
```
alexander-veit, line 42:
```
if qc.get("derived_from") == COVERAGE_DERIVED_FROM:
    return qc["value"]
```
might be cleaner

alexander-veit (approved):

PR #443, opened 2025-06-04
SN Update protected donor

sarahgonicholson:
- Embed `donors.protected_donor` on File
- Move `tissues` revlink from AbstractDonor to Donor
- Add embeds from MedicalHistory to ProtectedDonor


crfmc (approved):
Looks good to me!

PR #445, opened 2025-06-05
Data Matrix & Top Bar Navigation Updates

utku-ozturk:
- Resolves the issues in https://hms-dbmi.slack.com/archives/C051155U7QC/p1748988694063539

image

- Updates top bar navigation and adds `Data Overview` section under `Data` 

image


crfmc:
The nav bar items start to collide below ~1500px screen width (and below). I think we can decrease the font sizes, and then for mobile, switch to column alignment earlier.

Screenshot 2025-06-05 at 6 56 14 PM

crfmc:
I am noticing a potential inconsistency (not sure if intended behavior). The number under "Total Files" sometimes shows a different number than the search view total.

The count under "Total Files" in the data matrix seems to be exclusively that assay. If this difference is expected, maybe we can write "Browse ALL files" to clarify that users might expect a different total in the search result.

Screenshot 2025-06-05 at 7 09 01 PM
Screenshot 2025-06-05 at 7 28 09 PM

src/encoded/static/components/static-pages/components/matrix/StackedBlockVisual.js
```diff
@@ -1374,16 +1441,49 @@ const Block = React.memo(function Block(props){
                 placement="bottom"
                 overlay={popover}
                 rootClose
```
crfmc, line 1443: I pulled in the changes to test on Donor View. It looks great. Could we add a `flip={true}` here to prevent overflowing the bottom of the page?
```suggestion
                rootClose
                flip={true}
```

crfmc (approved):

PR #467, opened 2025-06-26
feat: update bulk donor metadata download button

crfmc:
[Trello Ticket](https://trello.com/c/KFq7KaIG/704-update-bulk-donor-metadata-link)

Update download link for bulk donor metadata download link


utku-ozturk (approved):
@crfmc  Looks good to me! My only concern is how we know whether the first search result item is the latest file without explicit sorting. (Since response has only one item, it is not blocking for now.)

PR #468, opened 2025-06-26
SN File QC notes

sarahgonicholson:
- Add `qc_notes` calc prop to QualityMetrics which uses `flag` to identify key metrics and build a concatenated string of Warn/Fail QC metrics
- Embed `quality_metrics.qc_notes` on File
- Add `quality_metrics.qc_notes` as column in File Manifest

### DIFF
--- a/src/encoded/metadata.py
+++ b/src/encoded/metadata.py
@@ -25,6 +25,7 @@ def includeme(config):
     config.scan(__name__)
 
 
+TSV_WIDTH = 27 # There are 27 columns in the file manifest
 # Encode manifest file types
 FILE = 0
 CLINICAL = 1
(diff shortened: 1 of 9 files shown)


alexander-veit (approved):

src/encoded/metadata.py
```diff
@@ -520,8 +523,8 @@ def generate_other_manifest_header(manifest_enum):
 
 
 def generate_file_download_header(download_file_name: str, cli=False):
-    """ Helper function that generates a suitable header for the File download, generating 26 columns"""
-    header1 = ['###', 'Metadata TSV Download', 'Column Count', '26'] + ([''] * 22)  # length 26
+    """ Helper function that generates a suitable header for the File download, generating 27 columns"""
+    header1 = ['###', 'Metadata TSV Download', 'Column Count', '27'] + ([''] * 23)  # length 27
```
aschroed, line 527: Use Constants for one or both of the numbers so only need to change in one place?

aschroed (approved):
Looks fine - I left one minor comment.

PR #470, opened 2025-06-30
Bulk file release

alexander-veit:
- Modified the release script to access multiple files to release
- Added a button to the Submission Status page, that will let you copy the accession of all output files displayed, so that it can be used in the release script.
- Minor adjustments to the QC overview script.


sarahgonicholson (approved):
Tested it out with some files on devtest and looks great!

PR #471, opened 2025-07-01
SN Update file merge group

sarahgonicholson:
- Update `generate_sample_source_part` in FileSet so that the sample source part of the `file_group` is Tissue, rather than TissueSample for single tissue samples with spatial information


willronchetti (approved):
Worth testing on devtest prior to release

PR #479, opened 2025-07-11
Restricted + Public Restricted

willronchetti:
- Adds two new ACLs for `restricted` and `public-restricted` that allows metadata items to be permissions via presence of the `dbgap` and `public-dbgap` groups
- Expands the `restricted` status to apply a global block on dbGaP permission for certain metadata items
- Allow the `public-restricted` status to function similarly to how `restricted` functions for non-file items ie: anyone can view, only protected users can download
- Tests for protected donor and medical history using the `restricted` and `public-restricted` statuses, along with tests for files as well utilizing the new groups

### DIFF
--- a/src/encoded/tests/test_permissions.py
+++ b/src/encoded/tests/test_permissions.py
@@ -1112,3 +1112,39 @@ def test_authenticated_user_can_delete_access_key(
         assert access_key["status"] == "current"
         patch_body = {"status": "deleted"}
         patch_item(user_app, patch_body, access_key["uuid"], status=403)
+
+
+@pytest.fixture
+def protected_donor(testapp, test_submission_center):
+    item = {
+        "uuid": "35dae4c5-c50e-4eb3-a240-4e62749eaa2b",
+        "submitted_id": "TEST_PROTECTED-DONOR_FEMALE",
+        "external_id": "SMHT001",
+        "submission_centers": [
+            test_submission_center['uuid']
+        ],
+        "age": 65,
+        "sex": "Female",
+        "hardy_scale": 3,
+        "tpc_submitted": "True",
+        "status": "restricted"
+    }
+    res = testapp.post_json('/ProtectedDonor', item)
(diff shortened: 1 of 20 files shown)


src/encoded/tests/data/master-inserts/static_section.json
```diff
@@ -534,7 +534,7 @@
             "default_open": true
         },
         "section_type": "Page Section",
-        "uuid": "68271697-2726-4bbf-9243-1c2ae9553c80",
+        "uuid": "27d09f44-44ca-466e-ade6-6d5a5a81c99c",
```
sarahgonicholson, line 537: In `master-inserts/page.json` line 1031, I think we need to change the `uuid` to "27d09f44-44ca-466e-ade6-6d5a5a81c99c" as well, otherwise it throws and insert error

src/encoded/tests/test_permissions.py
```diff
@@ -418,9 +419,34 @@ def test_dbgap_group_with_restricted_status(submission_center_user_app, submissi
         submission_center2_user_app.get(f'/{atid}@@download', status=403)
         anontestapp.get(f'/{atid}@@download', status=403)
         authenticated_testapp.get(f'/{atid}@@download', status=403)
+        smaht_public_dbgap_app.get(f'/{atid}@@download', status=403)  # public dbGaP user can't aaccess
```
alexander-veit, line 422: Add `smaht_public_dbgap_app.get(f'/{atid}', status=200)` ?
Can public dbgap view metadata?

willronchetti:
> This looks good to me. I did have a question about which items get the restricted and public-restricted ACLs. Pathology reports and histology images do not need them?

We're double checking this, thanks for flagging!

alexander-veit (approved):
I think that looks good. Is there a table that visualizes the permission settings? E.g.
"group: dbgap and item status: restricted" -> "view: yes, download: yes"?

PR #484, opened 2025-07-21
SN Donor metadata release

sarahgonicholson:
- Creates the script `commands/donor_metadata_release.py` which takes in one or multiple Donor accession IDs and releases the Donor item and all other donor metadata, tissue, and TPC-submitted tissue sample items associated with it
- Includes the constants `PUBLIC_DONOR_RELEASE_STATUS` (currently set to `released`, but will later be changed to `public`) and `PROTECTED_DONOR_RELEASE_STATUS` (currently set to `released`, but will later be changed to `public-restricted`) to ensure only Donor, Tissue, and TissueSample items are made publicly viewable, and ProtectedDonor and all other downstream donor metadata items are only visible if users have dbGaP access
- Run with the command `release-donor-metadata -e data --dry-run -d [Donor Accession ID]`
- Will fail if a non-Donor accession ID is provided and give a warning if the Donor item has a status that is not `in review`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -186,6 +186,7 @@ parse-structured-data= "encoded.commands.parse_structured_data:main"  # For dev/
 prepare-docker = "encoded.commands.prepare_template:prepare_docker_main"
 prepare-local-dev = "encoded.commands.prepare_template:prepare_local_dev_main"
 release-file = "encoded.commands.release_file:main"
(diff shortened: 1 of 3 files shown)


alexander-veit (approved):
I think that looks good. That was a lot of work, so thanks for that. 
It was a good decision keep the style of the release script. That make it easier for both of us to understand what's happening. At one point we could move all the common helper function into a separate script to avoid code duplication, but you don't need to do it know.

PR #496, opened 2025-07-28
Refactor Data Matrix and Update Styles

utku-ozturk:
Move DataMatrix components to the Viz folder, enhance tab styles, and simplify data aggregation processes. Adjust visual elements and update tests to reflect the redesigned data matrix popover.


src/encoded/static/components/viz/Matrix/StackedBlockVisual.js
```diff
@@ -539,182 +564,168 @@ export class StackedBlockVisual extends React.PureComponent {
...
-                    leftAxisKeys = _.map(sortedKeys, (key) =>
-                        key['groupingKey']
-                    );
-                } else { //sort by row labels
-                    if (sorting === 'asc') {
-                        leftAxisKeys.sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
-                    } else if (sorting === 'desc') {
-                        leftAxisKeys.sort((a, b) => b.toLowerCase().localeCompare(a.toLowerCase()));
-                    }
+                console.log('xxx sortedKeys:', sortedKeys);
```
crfmc, line 583: Just a note to remove console log before merging

src/encoded/static/components/viz/Matrix/StackedBlockVisual.js
crfmc, line 1: One note about the column/row totals, (@shannonehmsen may have a better idea for this). I think if users can't click a cell to open the popover they should look a little different than the ones that user can open the popover on. Maybe we can remove the border and background or something? Also I think the cursor should be the default (instead of the pointer) when hovering over un-clickable cells

crfmc (approved):
The changes look good to me! Just some small comments

PR #499, opened 2025-07-30
WF Fix ingesting identical items

wfeng19:
* When ingesting metadata through `submitr`, the portal will attempt to patch metadata for all items on the spreadsheet, even if the item and the patch are actually identical. This causes an error with released items in particular, because released items are not allowed to be modified
* Added a diff check to remove any items that are not being modified from StructuredData


sarahgonicholson (approved):
Looks good!

willronchetti (approved):
Looks good pending integrated testing

PR #505, opened 2025-08-04
Handle Aorta correctly in tissue classification table

alexander-veit:
- Treat Aorta as a special case in the tissue classification table, as GTEX only has Blood Vessel.


crfmc (approved):
Looks good to me!

PR #511, opened 2025-08-12
SN Sample Nomenclature doc

sarahgonicholson:
- Update `sample_nomenclature.rst` to have more up-to-date tables and include link to version 1.3 of the nomenclature document


crfmc (approved):
Looks great! Only thing I noticed (not from this PR) is that the header for table 3B ([here](https://github.com/smaht-dac/smaht-portal/pull/511/files#diff-bb3918caf31c19513929771056a20c4b9bfcc0c604b0f823fd3922291a08c281R562)) does not have a period at the end like the others. Would be good to add for consistency.

PR #684, opened 2026-06-02
Optimize Bulk Download

willronchetti:
* Optimize ``/metadata`` and ``/peek-metadata`` to handle thousands of files without upstream timeouts
* ``/metadata``: yield TSV rows from a generator instead of buffering the full manifest in memory; ES batch size is now constant regardless of file count
* ``/metadata``: top-level and sub-entity (sample/analyte/file_set) searches now bypass snovault's default-facet machinery and paginate via ES ``search_after`` (O(N) instead of O(N²) from/size), using a new ``execute_streaming_search`` primitive
* ``/metadata``: restrict ES ``_source`` to only the columns the TSV reads, dropping per-hit payload by ~10×
* ``/metadata``: pre-compile per-row field-path splits so each hit pays one dict lookup per column instead of repeated ``str.split`` calls
* ``/peek-metadata``: compute the file_size summary by streaming the matched docs (same path as ``/metadata``) and summing in Python, instead of issuing an ES ``stats`` aggregation that was blocking on slow-shard coordination
* ``/peek-metadata``: GET-style requests (search-filter URL params) now forward through snovault ``search()`` with the new ``skip_default_facets=true`` flag, preserving nested-field correctness while skipping the dozens of schema-default facet aggregations


dbmi-svc-checkmarx:
Logo
**Checkmarx One – Scan Summary & Details** – [554956f7\-c012\-4dd2\-aae5\-28d95045ddab](https://us.ast.checkmarx.net/projects/e9cf671f-a7fd-493f-900d-663a40a13bee/scans?id=554956f7-c012-4dd2-aae5-28d95045ddab&branch=new_image)


---
*Examples:*
`@Checkmarx rescan this PR`

pyproject.toml
```diff
@@ -114,6 +114,7 @@ WSGIProxy2 = "0.4.2"
 markdown = "^3.5.1"
 pygments = "^2.17.2"
 setuptools = "76.1.0"
+pip = "^26.1.1"
```
alexander-veit: Does this need to be here?

src/encoded/metadata.py
```diff
@@ -882,6 +900,303 @@ def descend_field(request, prop, field_names, cli=False):
...
+    """
+    es = request.registry[ELASTIC_SEARCH]
+    es_index = get_es_index(request, [type_param or 'File'])
+
+    # search_after needs a stable, unique sort. The user's sort_param is
+    # applied as the primary key when present; uuid acts as the unique
+    # tiebreaker so we never skip or duplicate documents at page boundaries.
+    sort_fields = []
+    if sort_param:
+        sort_fields.append({f'embedded.{sort_param}.raw': {'order': 'asc'}})
```
alexander-veit, line 982: Ok to default to "asc" here? Or should `sort_param` be parsed for a leading "-" and then map to "desc"?

pyproject.toml
```diff
@@ -42,7 +42,7 @@ certifi = ">=2021.5.30"
 chardet = "^5.0.0"
 codeguru-profiler-agent = "^1.2.4"
 colorama = "0.3.3"
-dcicsnovault = "^11.30.0"
+dcicsnovault = "11.30.1.0b1"
```
alexander-veit: Release stable version first?

alexander-veit (approved):

PR #702, opened 2026-07-06
perf: reduce redundant work in visualization.py aggregation views

willronchetti:
## Summary

Internal performance refactors for the aggregation views in `src/encoded/visualization.py`. These reduce redundant Python work in `bar_plot_chart` and `data_matrix_aggregations` **without changing response shape, values, or route behavior** as observed by callers. Based on a prior profiling investigation.

A new behavior-locking test module (`src/encoded/tests/test_visualization.py`) mocks the Elasticsearch round-trip (`perform_search_request` / `make_search_subreq`) with canned aggregation results and asserts exact output for each route. I also golden-diffed the old vs. new module on the same fixtures to confirm byte-identical output.

## Changes

**`bar_plot_chart`**
- **2c:** Compute `doc_count` once per bucket instead of `int(...)` twice. (The per-bucket `all_donors_ids` list is genuinely distinct per bucket and correctly retained — see note below.)
- **2d:** Merge the `include_meta_tissue_categories` extraction into the single pass over the top-level buckets, eliminating a second full traversal of the same bucket list.

**`data_matrix_aggregations`**
- **3d:** Precompute the extra-total result-field names once instead of re-scanning/re-validating `EXTRA_TOTAL_AGGREGATIONS` for every bucket; cache the bucket `key` and the nested sub-aggregation dict inside `format_bucket_result` (previously re-derived via repeated dict lookups on a potentially large tree).
- **3b:** Hoist the request-invariant coverage Painless-script definition (`SUM_DATA_GENERATION_SUMMARY_AGGREGATION_DEFINITION`) to module scope so the dict isn't rebuilt on every request. (ES itself caches compiled scripts by source string, so no app-level script cache is needed; the field-parameterized scripts in `get_es_value` are left untouched to preserve correctness.)
- **3f:** Precompute the composite (list) row-field list in the `flatten_values` path instead of re-testing `isinstance` for every field of every item.

## Issue 2a — deliberately not changed

The report flagged `all_donors_ids` `"size": 10000`. Checking the frontend call sites, the **full** donor-id list per bucket is consumed to build `donors.external_id` browse filter links:
- `viz/BarPlot/Chart.js` → `all_donor_ids: termObj.all_donors_ids`
- `viz/ChartDetailCursor/ChartDetailCursor.js` → `baseParams['donors.external_id'] = last?.all_donor_ids`
- `browse/components/FacetCharts.js` → `newDonorFilters.external_id = last?.all_donor_ids`

Reducing the size (or swapping to a plain cardinality agg) would silently drop donors from the click-through filter links, so the size is left unchanged.

## Performance

Measured Python post-processing (ES mocked, GC-controlled, min of repeated runs) on a synthetic **30×30** data matrix:

| Route | Before | After | Δ |
|---|---|---|---|
| `data_matrix_aggregations` post-processing | 0.528 ms/call | 0.457 ms/call | **-13.5%** |

`bar_plot_chart`'s wall-time is dominated by the (intentionally preserved) donor-id list serialization, so its measured end-to-end change is small; the value there is removing the redundant second bucket traversal and per-bucket duplicate conversion.

## Testing

`src/encoded/tests/test_visualization.py` — 5 tests covering `date_histogram_aggregations`, `bar_plot_chart` (two-level + tissue categories), and `data_matrix_aggregations` (nested + `flatten_values`). Each asserts the full response structure/values. All pass; old-vs-new golden diff confirms identical output.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dbmi-svc-checkmarx:
Logo
**Checkmarx One – Scan Summary & Details** – [d02439df\-c9a2\-4afe\-86e5\-ab9f69e81caa](https://us.ast.checkmarx.net/projects/e9cf671f-a7fd-493f-900d-663a40a13bee/scans?id=d02439df-c9a2-4afe-86e5-ab9f69e81caa&branch=fm%2Fsmaht-viz-opt-9k)


---
*Examples:*
`@Checkmarx rescan this PR`

willronchetti:
## Follow-up review (self-review + UI consumers + peek-metadata)

### 1. Self-review / correctness re-verification
Re-ran the old-vs-new **golden diff** (pre-refactor module vs this branch) over the base fixtures **plus 6 edge cases** — byte-identical output in every case, including key insertion order. Added the highest-value edge cases as permanent tests (commit `a623c21`):
- `bar_plot_chart`: empty `field_0` buckets with tissue categories requested (single-pass 2d path, 0 buckets); category buckets with a `null` key and with no category buckets.
- `data_matrix_aggregations`: buckets missing `total_coverage`/`donors` (3d `extract_bucket_counts` defaults); composite column **and** row keys in the `flatten_values` path (3f).

Suite is 9/9 passing, flake8 clean. No half-done work or cross-route inconsistency.

### 2. UI consumers — shape safety confirmed
Reviewed every consumer of the three routes (`chart-data-controller.js`, `Matrix/DataMatrix.js`, `StatisticsPageViewBody.js`, `BrowseDonorVizWrapper.js`, `BrowseSummaryStatController.js`, `BarPlot/Chart.js` + `ChartDetailCursor.js`, `RecentReleasesTimelineMatrix.js`).

**No consumer depends on `terms`/`counts`/`meta` dict key insertion order** — `Chart.js` re-sorts terms, `BrowseDonorVizWrapper` parses/range-matches keys, `DataMatrix` iterates arrays. The shape contracts they *do* rely on are all preserved here:
- `total` stays an object with `files`/`donors` (chart-data-controller, summary stats);
- each `terms` node keeps `{term, field, total, terms, all_donors_ids}` (Chart.js, BrowseDonorVizWrapper — note `all_donors_ids` is consumed as the click-through donor filter list, which is why size stays 10000);
- flatten payload keeps `data`/`row_totals`/`column_totals`/`counts`/`facets`/`filters` (DataMatrix);
- `date_histogram`'s `interval` stays an **array** with `{interval}_interval_{field}` agg keys (StatisticsPageViewBody).

**UI-side performance observations (out of scope for this backend PR, flagged for follow-up):**
- Browse load fans out **~5 POSTs to `/bar_plot_aggregations/`** (2 filtered+unfiltered from ChartDataController, 1 from BrowseSummaryStatsViewer, 2 from BrowseDonorVizWrapper) with **no shared client cache**; the `sample_summary.tissues`+totals request is fetched twice (BrowseSummaryStatsViewer ≈ BrowseDonorVizWrapper Effect 2).
- `BrowseDonorVizWrapper` Effect 1 has **no abort/stale-guard** and refetches on every `href` change — a race/last-write-wins correctness risk (ChartDataController aborts and DataMatrix uses a request-id guard; this one doesn't).
- `SubmissionStatsViewController` fires **3 parallel `/date_histogram_aggregations/`** requests and re-issues all 3 on any control change, no debounce beyond React batching.
- `DataMatrix` re-POSTs identical bodies on facet off/on toggles (no cache), though it's otherwise the best-behaved consumer.

### 3. `/peek-metadata/` investigation
`src/encoded/metadata.py` `peek_metadata` (~line 1298) is **already heavily optimized** and has no visualization.py-style redundant-Python-computation to remove:
- POST path (`_aggregate_metadata_file_size`) streams matching docs via `search_after`, source-filters to `file_size`/`extra_files`, and sums in Python — deliberately avoiding the ES stats-aggregation coordinator step that caused upstream 504s. The POST caller (`SelectAllAboveTableComponent.js`) reads only the `file_size`/`extra_files.file_size` facets it returns.
- Manifest generators are generator-based (constant memory) with set-based dedupe.

**One tradeoff worth a decision (mirrors the donor-ID case — not changed unilaterally):** the **GET path** (`_facets_via_search`) forwards to `snovault.search` and computes the *full* default facet set (~15–20 aggregations for File), but each GET caller reads exactly **one** facet — `BrowseDonor`/`BrowseProtectedDonor` file-size column reads only `file_size` (and appears to fire once per donor row), and `ProtectedDonorViewDataCards` `DonorDSAValue` reads only `type` facet `doc_count`. Narrowing the computed facets would speed these up but risks the intentionally type-agnostic GET contract (different callers read different facets), so it needs a product decision rather than a guess.

willronchetti:
## Follow-up: convert single-facet /peek-metadata GET callers to POST

Per the captain's decision (switch single-facet GET callers to POST rather than narrow the shared GET path). Exhaustive caller audit was done previously; two GET callers were candidates.

### Converted: `ProtectedDonorViewDataCards` DSA existence check
This standalone GET read only the `type` facet's `File` doc_count to decide "does this donor have any DSA files?", yet the GET computes the full default facet set (~15-20 aggregations).

The **existing streaming POST path cannot serve this** — `_build_metadata_es_filter` only supports type/accessions/status/uuid, not the DSA query's `data_type` multi-value, `dataset!=No value` negation, `donors.display_title`, `sample_summary.studies`. Reimplementing that filter construction would be substantial and fragile.

Instead I used snovault's `skip_default_facets` (added to snovault specifically for peek-metadata-style callers, previously unused here): a new POST mode where a body carrying `search_query_params` runs `snovault.search` with `skip_default_facets=true&limit=0` — reusing snovault's correct filter construction while computing **zero** facet aggregations — and returns `{'total': N}`. The frontend now POSTs and reads `total > 0`. Because the query filters `type=File`, `total` equals the old `type`-facet `File` doc_count, so the existence answer is identical.

- GET path untouched → its general-purpose `/search`-equivalent facet contract is preserved.
- New POST mode is keyed on `search_query_params`, which the existing file-size/manifest POST bodies (`accessions`/`type`) never carry — no collision.
- Tests in `test_metadata_peek.py` (5, mocked ES round-trip): POST count == old GET `type`-facet read; facet-free/`limit=0` controls can't be overridden by the caller; the real DSA filters (incl. the `dataset!=No value` negation) are forwarded; file-size/manifest POSTs are unaffected.

### Not converted (deliberate): `BrowseDonor` / `BrowseProtectedDonor` file-size column
Its single GET (`customColumnSearchHref`, `additional_facet=file_size`) feeds **four** table columns reading four different facets: `sample_summary.tissues` (incl. schema-derived `has_group_by`/`original_terms` grouping metadata), `assays.display_title`, `type` doc_count, and `file_size` sum. The streaming path can only sum/count — it cannot produce terms facets, and the tissue grouping metadata comes only from the schema facet definition (not reproducible via `additional_facet`). Splitting into up to four POSTs would multiply requests while still requiring `snovault.search` for three of them. It's a small per-donor (scoped) query, so the full-facet cost is modest. Leaving it as one GET is the right call.

utku-ozturk (approved):
looks good to me! data matrix tab caching is great.

PR #703, opened 2026-07-06
Fix column-header select-all checkbox to select all matching files in Recent Releases page

utku-ozturk:
- The `@type` column header checkbox on file tables (e.g. Recent Releases) was meant to mirror the "Select All" button's behavior, but it silently capped selection at a single page (`defaultPageSize`) worth of items instead of paginating through all matching results.
- Root cause: the checkbox was rendered inside a wrapping `<div>` as `colTitle`, which defeated `HeadersRow`'s auto-injection of the live search `context` (the mechanism only clones props into a `colTitle` whose root element is itself a function component). Without a real `context.total`, `SelectAllFilesButton`'s paginated fetch-all logic had nothing to page through and stopped after one request.
- Fix: introduced `SelectAllHeaderCheckbox`, a function component wrapping `SelectAllFilesButton`, used directly as `colTitle` so `HeadersRow` can inject the live context. This reuses the exact same `fetchAllPagesForSelection`/`handleSelectAll` logic as the "Select All" button — no duplicated select-all logic.
- Added a progress indicator (spinner + thin progress bar + tooltip) to the checkbox variant of `SelectAllFilesButton`, matching the feedback already shown by the button variant, so users don't wonder if the fetch is still running while all pages are being selected.


utku-ozturk:
[Clicking select all checkbox in column header is not working correctly in recent releases](https://trello.com/c/V003PmB9/918-clicking-select-all-checkbox-in-column-header-is-not-working-correctly-in-recent-releases)

dbmi-svc-checkmarx:
Logo
**Checkmarx One – Scan Summary & Details** – [44ff98ca\-0116\-4d39\-b2d5\-678fd9ef8e10](https://us.ast.checkmarx.net/projects/e9cf671f-a7fd-493f-900d-663a40a13bee/scans?id=44ff98ca-0116-4d39-b2d5-678fd9ef8e10&branch=utk_col_header_select_all)


---
*Examples:*
`@Checkmarx rescan this PR`

crfmc (approved):
Looks good! Thanks for fixing. We may want to move of the inline styles for the progress bar into SCSS if we ever need to overwrite, but fine to leave for now if you prefer to keep everything together.

PR #713, opened 2026-07-09
refactor: load donor browse row data progressively with a concurrency-limited queue

crfmc:
[Trello Ticket 912](https://trello.com/c/iK00056y/912-look-into-whether-it-is-possible-to-reorder-requests-on-browse-by-donor)

## Summary

- Add `DonorDataProvider` / `DonorDataCell` render-prop pattern so all four donor-row data columns (tissues, assays, file count, total file size) share a single `/peek-metadata/` fetch per donor rather than issuing separate per-column requests.
- A concurrency queue (max 4 in-flight) loads donors in display order so the first visible rows populate immediately; rows added by infinite scroll enqueue automatically.
- The request uses `skip_default_facets=true` + three explicit `additional_facet` params, reducing Elasticsearch aggregation work from the 21-facet File default set to exactly the 3 aggregations the columns read.

### DIFF
--- a/src/encoded/metadata.py
+++ b/src/encoded/metadata.py
@@ -1360,9 +1360,8 @@ def peek_metadata(context, request):
 
     # GET path — forward URL params verbatim to /search. Type-agnostic; the
     # caller gets the same facets `/search?<their-params>` would return,
-    # which is what the legacy fallback did. This preserves callers like
-    # ProtectedDonorViewDataCards.js that read default facets (e.g. `type`)
-    # from the response.
+    # which is what the legacy fallback did. Callers that use
+    # `skip_default_facets=true` must explicitly request every facet they read.
     if isinstance(args, Response):
         return _facets_via_search(request, request.params)
 
--- a/src/encoded/metadata.py
+++ b/src/encoded/metadata.py
@@ -1045,7 +1045,7 @@ def _facets_via_search(request, params):
     aggregation coordination timeout). Everything else goes through here.
     """
     forwarded = MultiDict()
-    for key in params.keys():
+    for key in dict.fromkeys(params.keys()):
         if key in ('limit', 'from'):
             continue
         for value in params.getall(key):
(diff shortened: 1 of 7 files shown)


dbmi-svc-checkmarx:
Logo
**Checkmarx One – Scan Summary & Details** – [0a1e12cf\-0139\-432f\-a704\-5ea801e300ef](https://us.ast.checkmarx.net/projects/e9cf671f-a7fd-493f-900d-663a40a13bee/scans?id=0a1e12cf-0139-432f-a704-5ea801e300ef&branch=cfm-donor_browse)


---
*Examples:*
`@Checkmarx rescan this PR`

src/encoded/metadata.py
utku-ozturk, line 1: In the local environment, the `/peek-metadata` response appears to be repeating the required fields three times. I don't think this is within the scope of your PR, but it may be worth taking a quick look before merging.

Image

src/encoded/static/components/browse/browse-view/BrowseProtectedDonor.js
```diff
@@ -161,7 +156,7 @@ export const BrowseProtectedDonorBody = (props) => {
     const { context, alerts, href, userDownloadAccess, isAccessResolved } =
         props;
     return (
-        <>
+        <DonorDataProvider>
```
utku-ozturk: I think `DonorDataProvider` needs invalidation when the browse state changes. Right now it is mounted once around the whole donor browse page, and its cacheRef / pendingRef / queueRef persist across filter, sort, and pagination changes. If the user changes the result set while requests are still queued, the provider keeps draining the old queue first, so newly visible rows can stay in loading while concurrency is spent on donors that are no longer on screen. That seems to break the intended “populate in display order” behavior for the current view.

One simple fix (for both `BrowseDonor` and `BrowseProtectedDonor`) would be to remount the provider when the browse URL changes, for example by keying it on `href`, e.g.:

``` jsx
<DonorDataProvider key={href}>
    ...
</DonorDataProvider>
```

crfmc: Great catch! Thank you for the suggestion

PR #721, opened 2026-07-18
fix: harden nginx failover/observability (defer rss_limit raise)

willronchetti:
## Summary

Hardens nginx failover and observability for waitress worker SIGKILL/restart churn. The per-worker `rss_limit` increase proposed in the first revision is deferred: this PR leaves the cap at **450MB**.

The branch includes merged `main` through PR 722. Release metadata is reconciled as **2.4.2** for PR 721, retaining PR 722 under 2.4.1 in the changelog.

## Capacity boundary

The five `snovault.memlimit` checks cover Python parent RSS; they do not form a task-wide bound because thread-local Node SSR children, nginx buffers, and watchdog overshoot are outside those caps. Raising every parent from 450MB to 600MB would permit another 750MB of parent RSS in an 8 GiB task. A future increase therefore needs a same-window process-tree/cgroup budget, canary alarms, and a defined rollback threshold.

Making port 6547 active does warm the fifth worker's lazy Node SSR children and can raise task RSS even at the unchanged parent cap. This PR does not claim that the unchanged cap bounds total task memory.

## nginx changes

- Makes all five workers active (`6547` was `backup`) so normal traffic round-robins across five parents instead of concentrating on four.
- Reduces `fail_timeout` from 45s to 15s so a recovered peer becomes selectable sooner. `max_fails` remains at its default of 1. Supervisord's `startsecs=6` is a liveness window, not a measured socket-readiness time.
- Sets `proxy_next_upstream_tries 2`, allowing at most **two total upstream attempts**, and `proxy_next_upstream_timeout 30s`, which limits when another handoff may begin rather than imposing an end-to-end deadline.
- Keeps nginx's default method protection: POST/LOCK/PATCH are not retried after being sent upstream; pre-send connection failures may retry any method, and PUT/DELETE/OPTIONS and other unprotected methods may retry.
- Adds an `upstream_debug` stream for multiple-attempt failover and upstream 502/504 gateway failures. It logs method, normalized URI without the query string, nginx request ID, timing, and upstream addresses/statuses. It omits the client address, and ordinary application 5xx responses remain solely in the LB log.

## Deliberate limits

- A request that selects two failed peers can return an error even if a third peer is healthy. This bounds retry amplification and latency during correlated failures.
- A worker failure after response bytes reach the client cannot be retried; the client receives a truncated/failed response.
- Non-protected methods such as PUT and DELETE may be replayed once. The policy does not enable nginx's `non_idempotent` option.
- The targeted nginx stream is separate from the LB log, so its retention and access controls must match.

## Validation

- The production Docker build runs `nginx -v && nginx -t` after installing the checked-in config, gating syntax and directive compatibility against the pinned nginx 1.21.6.
- `python3 deploy/docker/production/test_nginx_failover.py`: **10/10 pass** locally on nginx 1.27.0. The offline harness checks the checked-in 450MB/five-peer/retry/logging guardrails; GET failover; POST after-send no-retry; POST pre-send retry; PUT retry; the two-attempt limit; query/client-address minimization; exclusion of ordinary application 500s; and inclusion of upstream 502/504 signals.
- Merged-main Sentry regression suite: **7 passed** offline; focused flake8 is clean.
- Python compilation, `git diff --check`, release-version parsing, and conflict-marker checks pass. `pyproject.toml` and the top changelog entry are both 2.4.2.

The prior UNIT run's two setup errors were traced to one shared remote OpenSearch create request: the initial `sequencing` index `PUT` timed out after 20 seconds but completed server-side, then the client's retry received `resource_already_exists`. All 1,240 executed tests passed before those fixture errors, and the cleanup step ran. No nginx/application workaround is included for that transient remote-test condition; this merged commit triggers a fresh isolated CI run.

No deployment, AWS, production-service, or live-Sentry action is included.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,33 @@ Change Log
 ----------
 
 
+2.3.10
+======
+
+Production memory/reliability tuning for the per-worker RSS watchdog and nginx failover.
+
+* Raise the production per-worker ``rss_limit`` from ``450MB`` to ``600MB`` in ``base.ini``.
+  The in-process ``snovault.memlimit`` watchdog SIGKILLs a waitress worker when its RSS crosses
+  this cap; at 450MB workers were being killed at ~474 MiB while the 8 GiB ECS task ran only
+  ~60% typical / <90% peak, wasting most of the task budget. 5 workers x 600MB (~3 GiB) leaves
+  ample headroom. Byte-limit behavior is preserved; no percentage limit is added because
+  ``psutil.memory_percent`` reads the Fargate host, not the task. This is a conservative capacity
+  adjustment, not a memory-leak fix. Rollback: revert the value to ``450MB``.
+* ``deploy/docker/production/nginx.conf`` reliability/observability changes for worker
+  SIGKILL/restart churn:
+
+  * Make all 5 workers active in ``upstream app`` (port 6547 was ``backup``, concentrating normal
+    traffic on 4 workers); failover is preserved by ``proxy_next_upstream``.
+  * Reduce upstream ``fail_timeout`` from ``45`` to ``15s`` so a worker restarted by supervisord
+    (~6s) is not quarantined for 45s; ``max_fails`` left at its default.
+  * Bound retries with ``proxy_next_upstream_tries 2`` and ``proxy_next_upstream_timeout 30s`` to
+    prevent a correlated kill storm from stampeding all upstreams. Non-idempotent requests are
+    still not retried.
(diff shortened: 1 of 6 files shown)


dbmi-svc-checkmarx:
Logo
**Checkmarx One – Scan Summary & Details** – [db4cced9\-78c0\-4825\-87e4\-3a5de97f8c16](https://us.ast.checkmarx.net/projects/e9cf671f-a7fd-493f-900d-663a40a13bee/scans?id=db4cced9-78c0-4825-87e4-3a5de97f8c16&branch=smaht-rss-memlimit-nginx-reliability)


---
*Examples:*
`@Checkmarx rescan this PR`

PR #728, opened 2026-07-20
260713 ajs fixed to frozen

aschroed:


dbmi-svc-checkmarx:
Logo
**Checkmarx One – Scan Summary & Details** – [5c39f5e5\-2306\-4575\-858b\-76a7a454de40](https://us.ast.checkmarx.net/projects/e9cf671f-a7fd-493f-900d-663a40a13bee/scans?id=5c39f5e5-2306-4575-858b-76a7a454de40&branch=260713_ajs_fixed_to_frozen)


---
*Examples:*
`@Checkmarx rescan this PR`

src/encoded/commands/associate_fixed_samples.py
```diff
@@ -0,0 +1,607 @@
...
+            fixed_protocol = FRESH_TO_FIXED_PROTOCOL_MAP.get(fresh_protocol)
+            fixed_group = fixed_index.get((donor_uuid, fixed_protocol), [])
+            if not fixed_group:
+                self.add_warning(
+                    f"No TPC fixed counterpart (protocol {fixed_protocol}) for donor "
+                    f"{donor_uuid}, GCC fresh protocol {fresh_protocol} "
+                    f"({len(fresh_group)} fresh sample(s)). This can be expected if "
+                    f"TPC hasn't submitted/processed this block's fixed samples yet."
+                )
+                continue
```
willronchetti: Stale association reconciliation: when a fresh sample's donor/protocol group has no current fixed counterpart (this branch), the command only warns and leaves any existing `linked_fixed_samples` on the record untouched. `compute_and_delete` (and the `--identifiers` scope generally) also only reaches samples still eligible under today's fresh/frozen + non-TPC selectors, so a record that has since become ineligible (its fixed counterpart deleted, or its own preservation/center changed) can't be reached to clean up a stale link either.

Could you clarify the intended contract: is `linked_fixed_samples` meant to be a point-in-time snapshot (please document that drift explicitly if so), or should it always reflect current reality? If the latter, reconciliation and `--delete` need to clear stale links even when the desired set is empty or the source/target no longer match current selectors, not just warn.

src/encoded/schemas/tissue_sample.json
```diff
@@ -100,6 +100,16 @@
...
+        "linked_fixed_samples": {
+            "title": "Linked Fixed Samples",
+            "description": "Fixed TissueSamples from the same tissue block as this fresh/frozen sample.",
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "type": "string",
+                "linkTo": "TissueSample"
+            }
+        },
```
willronchetti, line 113: `linked_fixed_samples` is documented as a derived, command-managed relationship, but nothing enforces that: the schema only requires `linkTo: TissueSample` (no `permission: restricted_fields`), and the existing add/edit validators in `types/tissue_sample.py` don't inspect this field at all. As written, any editor with edit permission on a fresh/frozen sample can PATCH it to point at an unrelated donor's fixed sample (or even another fresh sample), and `associated_pathology_reports` will serialize whatever they chose.

Recommend marking this field operator-restricted (`restricted_fields`) and adding a link-related validator that checks source eligibility (GCC, fresh/frozen) and target invariants (non-deleted status, NDRI-TPC center, `Fixed` preservation, matching donor, mapped protocol pair) rather than accepting any TissueSample link.

src/encoded/commands/associate_fixed_samples.py
```diff
@@ -0,0 +1,607 @@
...
+    ) -> None:
+        current = sorted(
+            item.get("uuid", "") if isinstance(item, dict) else item
+            for item in (fresh_sample.get(LINKED_FIXED_SAMPLES_FIELD) or [])
+        )
+        identifier = self._get_identifier_to_report(fresh_sample)
+        if current == fixed_uuids:
+            self.patch_infos.append(
+                f"  - {identifier}: already up to date ({len(fixed_uuids)} linked)."
+            )
```
willronchetti, line 334: `current` here reads whatever's already stored in `linked_fixed_samples` (dicts with a `uuid` key from search/embedded frames, but bare resource-path strings when sourced via the `frame="object"` `RequestHandler` used in `--identifiers` scope), then compares it directly against `fixed_uuids`, which are bare UUIDs. Since the object frame canonicalizes stored UUID links to resource paths, a targeted `--identifiers` run against an already-correct record will see `['/tissue-samples/<uuid>/']` vs. desired `['<uuid>']`, never match, and re-PATCH (and reindex) a semantically unchanged sample every time -- contradicting the "already up to date" message just above.

Recommend normalizing every current-link representation (dict UUID, bare UUID, or resource path/@id) through one helper before comparing, and covering all three representations in tests.

src/encoded/commands/associate_fixed_samples.py
```diff
@@ -0,0 +1,607 @@
...
+    def _check_for_orphaned_fixed_samples(
+        self,
+        fresh_by_donor_protocol: Dict[Tuple[str, str], List[Dict[str, Any]]],
+        fixed_index: Dict[Tuple[str, str], List[Dict[str, Any]]],
+    ) -> None:
+        # NB: 1J and 1K both map to fixed protocol 1L, so this reverse map
+        # collapses that pair -- purely cosmetic (it only affects which fresh
+        # protocol code gets named in the warning text below), not a
+        # correctness issue for the forward matching above.
+        fixed_to_fresh_protocol = {v: k for k, v in FRESH_TO_FIXED_PROTOCOL_MAP.items()}
```
willronchetti: Both `1J` and `1K` map to fixed protocol `1L` in `FRESH_TO_FIXED_PROTOCOL_MAP`, but this `{v: k for k, v in ...}` comprehension only keeps one of them (`1K`, since it's the later entry and overwrites `1J`'s reverse mapping). The comment above calls this cosmetic, but it does affect this function directly: a donor with a valid `1J` fresh group and a matching `1L` fixed group has no `1K` fresh group, so `_check_for_orphaned_fixed_samples` reports those `1L` fixed samples as orphaned even though they're correctly linked -- a false-positive warning in the dry-run safety gate.

Recommend reverse-mapping fixed protocols to a *set* of fresh protocols, and treating a fixed group as covered if any of its mapped fresh protocols is present.

willronchetti:
Test coverage: this PR doesn't touch any test files, and there's no coverage yet for the new `linked_fixed_samples` / `associated_pathology_reports` data model or the `associate-fixed-samples` command.

Could you add focused unit and integration coverage for: protocol mapping (including the `1J`/`1K` -> `1L` case), source/target filtering, reconciliation and cleanup of stale links, idempotency across repeated runs, historical records that lack the field, wrong target/status/donor/protocol combinations, and indexed calculated-property (ES) behavior for both `pathology_reports` and `associated_pathology_reports`?

src/encoded/commands/associate_fixed_samples.py
aschroed: Added `_normalize_link_identifier` to address.

src/encoded/schemas/tissue_sample.json
aschroed: Added both the `restrict_fields` permission and the validators `get_fixed_to_fresh_protocols...`

src/encoded/commands/associate_fixed_samples.py
aschroed: script updated to manage this and clean up 'stale' links as necessary.

src/encoded/commands/associate_fixed_samples.py
aschroed: `get_fixed_to_fresh_protocols` reverses the lookup to resolve this (albeit this only applied to benchmarking where may not be relevant as we don't have path reports for those and may never have.

aschroed:
Added tests for calcprops and validators and added functjons to item_utils but skipping tests for script as will be 'live' tested and will eventually be converted to foursight script.

willronchetti (approved):
Latest changes look good

utku-ozturk (approved):
looks good to me!

